Issue #22, opened 2015-11-23
App Icons


brandondean:

ZachOrr:
https://github.com/the-blue-alliance/the-blue-alliance-ios/commit/11a050823cf87bc3da14329dacaea9c33dc9b41a

PR #24, opened 2015-11-24
App Icons

brandondean:
Nothing really special here. Using the blue that's in the app with the lamp logo. #22 
screen shot 2015-11-23 at 23 43 43 pm


nwalters512:
If you want something that could fit in better with Apple's flat icons, you could use the flattened lamp from the Android app: https://github.com/the-blue-alliance/the-blue-alliance-logo/blob/master/android/svg/ic_launcher_blue.svg

brandondean:
@nwalters512 I would prefer a flat logo like Android, but I didn't want to go against branding. We shall see what happens in the-blue-alliance/the-blue-alliance#1285

synth3tk:
For the [Windows app](https://github.com/the-blue-alliance/the-blue-alliance-windows/tree/mvp/TheBlueAlliance/Assets), I've also simply adapted the Android icons, although I had to scale them up and down (thank goodness for SVG).

I think that's the general direction we're going in for the short-term until that issue is resolved.

ZachOrr:
Merging this for now so we can have an app icon

PR #25, opened 2015-12-29
Remove back button title text

ZachOrr:
To help with text clipping (either the back button text, or the navigation bar title) this sets up a delegate object (retained by the app delegate) to modify the back button item to be a back button without text


ZachOrr:
Sorry everyone, merged this one myself, mostly for the workspace change 😅We can revert if necessary, it's not much code.

PR #31, opened 2016-02-24
Nuke Settings and MyTBA

brandondean:
In an attempt to MVP this app out let's just nuke Settings and MyTBA and not worry about developing them for now.

Of course we'll revisit this and add it to the app, but for now let's focus on delivering something.

simulator screen shot feb 23 2016 20 16 22 pm


ZachOrr:
Let's do it :+1:

PR #33, opened 2016-03-22
Added a target to the Podfile to be compatible with CocoaPods 1.0 syntax

DomenicP:
The app dependencies were not installing using newer versions of CocoaPods (1.0.0.beta.5) and Ruby (2.3.0) because the Podfile was still using the pre-1.0 syntax. I updated the Podfile to include the proper target as is now required.


ZachOrr:
Hey Domenic! Sorry it took me so long to respond. I promise I looked at this sooner than today.

For now I'm going to keep the syntax for the current stable Cocoapods version and not the beta version. Once 1.0 becomes the stable version, we'll go ahead and update the Podfile syntax.

PR #34, opened 2016-04-23
Modifies data layer

ZachOrr:
Trying to do some abstracting to make things clearer and not repeat code as much. The big notes here is

1) New model superclasses to manage single insertions
2) Once-over'd the data model for nullability/default things
3) Actually make use of a main and private context for FRCs and background imports

Additionally I'm sneaking in a few things

1) Observing context changes in view controllers without fetch results controllers to update table views when their backing individual models are updated
2) Convert years to NSNumbers across the app to make use of type safety (and remove a lot of type casting)
3) Removes all the unused +Fetch classes (the other ones still kill me inside, but for now they're fine)
4) Added `setEvents:forYear:` method to the Team model in order to properly update events with the call we make in the EventsViewController - I'm sure there are some other classes that can use a method like this for updating one-to-many relationships for year-specific data
5) Remove a lot of reload/save code that we should be letting happen automatically

With this parent-child Core Data setup I'm thinking we might be able to lean on Core Data's unique constraints to ensure even more safety for multiple objects. Since the unique constraint resolution doesn't take place until saves, but we have to save the child context before saving the parent context, we'd never see duplicates in the UI. I'll have to investigate how setting multiple unique constraints works. My guess is it won't be powerful enough for some of the situations we have and we'll have to keep leaning on using the predicates for only finding one object.

The LAST big thing I need to tackle is some issues with the initial load. We do a lot of importing on the background context, which we propagate to the main context, but it takes a while and our loading indicator goes away before the save on the main context is done. My guess is that I can just set a callback for that `performChanges:` method and remove the spinner after that... but that's kinda hacky, since the spinner right now only indicates network requests and not data saves. I might be able to take a pass at batching things to speed them up, or maybe a third context that only handles writing data on the main context?


ZachOrr:
Aah got it - it ended up being this infinite loop bug with inserting events on the background context and trying to fetch them on the main context, then seeing that they weren't on the main context (yet, the save hadn't finished) and trying to fetch them on the background context, and so on.

phil-lopreiato:
> The LAST big thing

Woo! Almost there!

ZachOrr:
"The last big thing" for that pull request 😝






On Sat, Apr 23, 2016 at 7:06 PM -0700, "Phil Lopreiato" <notifications@github.com> wrote:













The LAST big thing




Woo! Almost there!

PR #49, opened 2016-04-30
[Closes #36] Add team at event stats

### LINKED ISSUE #36: Add Stats view to Team @ Event
Should be fairly straight forward - we can use the SummaryCell 
screenshot_20160429-220025

ZachOrr:
A little out of scope - also removed "unnecessary" (I think we don't need them, or at least we shouldn't be calling them so I'll fix it if it doesn't already remove them) hideNoDataView calls from refresh methods


brandondean:
LGTM

PR #57, opened 2016-05-01
Show play button for matches that have videos.

### LINKED ISSUE #56: Add Play button to match cells that have videos

brandondean:
Puts the match label and play button in their own vertical stack view, and
hides the play stack view if there isn't a video available for the match.

Closes #56
simulator screen shot may 1 2016 18 20 49 pm


ZachOrr:
Is there any way we could grab the same play button the Android app uses?

brandondean:
simulator screen shot may 1 2016 19 07 09 pm

ZachOrr:
Made a few changes to make sure that the play button and two lines of text will always fit properly

phil-lopreiato:
Heads up, the Android app gets most of its icons here (some are a little outdated/less material-ey): https://design.google.com/icons/

PR #59, opened 2016-05-02
Modify info view headers

ZachOrr:
Looking for feedback - might need to take another pass at how the Team view looks since there's more info there. But I think the Event view is pretty good.
simulator screen shot may 2 2016 1 11 22 pm
simulator screen shot may 2 2016 1 11 27 pm

### DIFF
--- a/the-blue-alliance-ios/EventsStoryboard.storyboard
+++ b/the-blue-alliance-ios/EventsStoryboard.storyboard
@@ -1008,7 +1008,7 @@
                                     <rect key="frame" x="0.0" y="0.0" width="600" height="43.5"/>
                                     <autoresizingMask key="autoresizingMask"/>
                                     <subviews>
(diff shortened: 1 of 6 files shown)


ZachOrr:
Gonna merge this - hopefully we can find a way to bring back in the sponsors at some point

brandondean:
Seems reasonable to me. Maybe another section underneath the name for the sponsors? It'll visually disconnect it, keeping the nickname obvious, but it might push down the social section just a bit too far.

ZachOrr:
Another section might be a good call. I thought about making it another cell in the first section. Or even throwing a cell underneath the team info header with the title "Sponsors" and a disclosure indicator. I don't know how interesting sponsors are to most people, and it does take up quite a bit of screen real estate.

brandondean:
simulator screen shot may 2 2016 19 37 46 pm

brandondean:
Meh. Could shrink it like android does. Limiting it to 2 lines, and then expanding it on tap.

ZachOrr:
I'd be okay with that. Expanding cells are a little weird, but in a lot of cases sponsors are just going to be a big wall of text. So any way of hiding it is probably a good call.

PR #61, opened 2016-05-03
[Closes #55] Setup initial background fetch for all teams

### LINKED ISSUE #55: Background fetch Team objects
I removed a lot of (gross) code for synchronously fetching a team object when we didn't have it and instead create stub teams to work on in ec7d23c24c046fbfa91ef23f4ec158663edb61a5. This worked out well from a data standpoint, but is very confusing form a UI standpoint.

My thought is we could approach this a few different ways. We could add back in the synchronous fetching code, we could do a fetch for teams periodically on launch to make sure they're up to date, or (my favorite option) we could create a background task that gets fired off when creating a stub Team object to fetch the rest of the data and resolve the missing data for the object.

This might be a bit of work - we need to make sure team objects will update in their contexts when the object is changed. I believe we'll get this for free anywhere we're using a FRC, but we need to make sure in places we're not using a FRC but are using team data (like the Info view) we register for changes.

ZachOrr:
After building a downloader that hooked in to inserts for stub team objects, I realized it was way more complicated than it needed to be, and making far too many network requests. I suggest we just attempt to refresh teams on initial launch to fix the current stub team states. We could also set a last fetched date in NSUserDefaults and if we see we haven't done a batch refresh in a certain period to keep things in sync.

### DIFF
--- a/the-blue-alliance-ios/AppDelegate.m
+++ b/the-blue-alliance-ios/AppDelegate.m
@@ -11,7 +11,6 @@
 #import "TBAViewController.h"
 #import "TBANavigationController.h"
 #import "TBANavigationControllerDelegate.h"
-#import "Team+Fetch.h"
 #import "OpenInGoogleMapsController.h"
 
 @interface AppDelegate ()
@@ -33,9 +32,6 @@
         UIStoryboard *storyboard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
         UITabBarController *rootTabBarController = [storyboard instantiateViewControllerWithIdentifier:@"RootTabBarController"];
         self.navigationDelegate = [[TBANavigationControllerDelegate alloc] init];
-     
-        // Refresh all our teams, if we don't have any
-        [self setupTeamsForPersistenceController:self.persistenceController];
         
         // Pass persistence controller to all navigation controllers
         for (TBANavigationController *nav in rootTabBarController.viewControllers) {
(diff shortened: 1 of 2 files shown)


ZachOrr:
Or, we could just fetch during every launch - not sure if that's a bad idea or not

brandondean:
If only we had a true paged team list API that could fix this issue pretty easily.

ZachOrr:
What's wrong with the current paged endpoint? Besides the lack of usual paging params (limit, offset, etc.)
I'm sure we could probably ask for more filtering in v3 😁 Particularly getting all teams after a certain number, or passing a list of teams we already have






On Mon, May 2, 2016 at 7:57 PM -0700, "Brandon Dean" <notifications@github.com> wrote:












If only we had a true paged team list API that could fix this issue pretty easily.

brandondean:
Oh snap, you are right. 

Why don't we fetch the first 500 teams when you visit the team list (the first time), and then page from there? Every 500 teams you scroll you simply have to wait for a refresh. With the team list doing this, and the team list for teams at events you will slowly fill the database with teams, but only with teams you visit.

brandondean:
I guess I don't really see the point of doing this in the background, and fetching teams that the user may never even visit.

ZachOrr:
The problem is with loading events. Events is the first call you make, and it returns event objects which have alliance arrays that contain team keys. In order to display details on the team @ event view for where a team was picked (or if they weren't picked at all) we have to establish a relationship between these alliance arrays/team keys and team objects. But, when making the events call, the only piece of data we have is team keys, so I defaulted to creating stub objects.

Fetching teams when making events calls isn't a great idea, since we'd be making team requests for every team that got picked at any event for an entire year, which is a lot of calls (and a lot of duplicate unnecessary calls).

After creating stub team objects, you can get in to states in the UI where you're missing a lot of data you'd be expecting to see. For instance, the teams view, or the rankings view at an event (screenshots attached).

I'm up for finding other solutions for approaching this problem 
simulator screen shot may 2 2016 11 18 48 pm
simulator screen shot may 2 2016 11 18 56 pm

brandondean:
> Fetching teams when making events calls isn't a great idea, since we'd be making team requests for every team that got picked at any event for an entire year, which is a lot of calls (and a lot of duplicate unnecessary calls).

I guess this where the breakdown is. I don't see the problem with making this request (for teams at event, in addition to the initial event call) since the request is fairly cheap IMO.

ZachOrr:
Sounds like a plan!

brandondean:
Ultimately it would be nice if the event returned team objects for everyone at an event.

I think I made this call way back when first writing this API, and I didn't really know what I was doing (and still don't).

ZachOrr:
Alright, how about something like this. The Teams view gets a little jumpy if you tab over while the inserts are still saving, but it's not the worst thing ever.

brandondean:
This works for me.

PR #62, opened 2016-05-03
Removes the Google Maps library

ZachOrr:
We no longer need it since we took out the action of opening an event/team location in Google Maps


ZachOrr:
Did this get merged? Or just closed? 😅

brandondean:
OH SNAP!

Good catch.

Issue #10, opened 2014-05-25
Contributors List?


donald-pinckney:
Should we plan on including a list of contributors, like the Android team is:
https://github.com/the-blue-alliance/the-blue-alliance-android/issues/32

PR #64, opened 2016-05-03
Add licenses of libraries we use.

brandondean:
Refs #10


ZachOrr:
Woooooaaaahhhh this is so cool 🙌

PR #65, opened 2016-05-03
Fix team pages refresh

ZachOrr:
When paging through teams with the Last-Modified header, we won't get any data back if a page for the teams endpoint hasn't been modified. However, the only way for us to tell if we've hit a page that doesn't have any teams is if we get back no data (an empty page still gives us a 200 and a Last-Modified back). This is a simple fix, but we store the max page we see for the teams paging call, and make sure we always check up to at least that point.


brandondean:
Rock'n roll!

PR #66, opened 2016-05-04
Split info views in to two, add expanding sponsors to Teams (Closes #60)

### LINKED ISSUE #60: Explore other options for showing team sponsors
Per #59 team sponsors got nuked from team summary view which isn't a bad thing since most of the time it's a wall of text, but we could figure out a way display the information without it hogging screen space.

Some options include:
* Put the sponsors in a cell on info view, limit the lines to something reasonable (like 2 or 3), and then resize expand the cell when tapped. This is what android does.
* Create a 'Sponsors' cell that is a disclosure into a screen that is just a giant text view. 😎

ZachOrr:
Info view was getting a little heavy with all the if statements with table view logic, so I split them in to two. We've gotta share similar cells across two view controllers now, but one is a basic cell and the other one is just a stack view.

Also, expanding sponsors! Thoughts?


fangeugene:
We have expanding sponsors on Android!

ZachOrr:
Yeah, I basically copied what Android is doing 😛

brandondean:
ah nice! Split the InfoViewController out for events too. 💅  it.

PR #79, opened 2017-05-12
added fancy graphics

rytonbay:
Hey there, I'm from team 1678, a graduating senior and lead of the mobile app programming subteam. I'd be really interested in helping push this dev project forward this summer, as TBA has been an incredibly useful tool for me during my 4 years as an FRC student, and I'd love to give back to the community a little bit. I made a few small additions in this PR, adding a simple splash screen, app icon, display name, and icons to the tab bar (non-intrusive changes while I get to know the code base). Thanks!


ZachOrr:
Hey Bryton - I appreciate your enthusiasm for supporting TBA's iOS app development efforts!

We already have [an app icon and some default tab bar icons](https://github.com/the-blue-alliance/the-blue-alliance-ios/tree/add-google-cloud-platform/the-blue-alliance-ios/Images.xcassets) picked out in the Obj-C branch that just haven't been moved over yet. We want to use these assets specifically as they're consistent with the visual design of the Android app.

If you could pull over those assets from the Obj-C branch and use them for this PR, that'd be great!

Here's a screenshot of the icons were using for each tab in the old branch

simulator screen shot may 11 2017 11 12 02 pm

rytonbay:
Hey there Zach, 

Well, looks like that was a waste of time. Too bad I didn't look closely at the objective-c branch. How much more complete is that version of the app? Why are we rewriting things in swift instead of just doing a combined language project where all new development is done in swift but we keep working objective-c code running? I trust that there are good reasons for all of this, I just need a little background to make sure I understand the general direction we want this project to go and to avoid more blunders like today's. Sorry to take your time on basic stuff like this. 

Also, I've CC'd Janet, another iOS programmer from my team who is interested in helping out with this. 

Thanks!
Bryton

ZachOrr:
No worries! These are all good questions.

I'd say the `add-google-cloud-platform` branch (the most recent branch I was working off of for the Obj-C codebase, which adds MyTBA functionality, but also a lot of other things) is probably 80% of the way "done", but the other 20% is in the details.

There are a lot of reasons for a Swift rewrite... bear with me

The Swift rewrite wasn't a light decision. We'd been hearing from potential contributors that they didn't know Objective-C and were hesitant to get involved. My stance has always been iOS work is mostly knowing the frameworks, but for less experienced programmers picking up unfamiliar syntax can be daunting. Since Swift/Obj-C doesn't matter to me, I weighted all the other factors in making the switch and decided we should go for it.

If you attempt to compile the Objective-C codebase you'll notice a handful of things are broken - but mostly Events. A little bit of [code that got added](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/Models/Event.m#L352) doesn't age super well and would probably get rejected nowadays if I saw it in a PR. Additionally some of [the data modeling I was doing](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/Models/Event.m#L150) in order to sort events was going to have to be rethought, since the initial design was neither eloquent or scalable. This would require rewriting a lot of the Event model, and the view controllers that consume Events.

The biggest influencer for me was APIv3, which we're currently using the native apps to consume/test. APIv3 was designed with mobile in mind, meaning we had a lot of say in how data was structured in order to make our lives easier. Ex: The Team@Event status view in Android was being generated entirely client-side, which is a lot of work. In APIv3 the server will figure this all out for us and return what we need. The [Obj-C version of TBAKit](https://github.com/ZachOrr/TBAKit/tree/changes) (the API wrapper I wrote for TBA) was all v2, and upgrading to v3 *and* moving to Swift (going from reference types to value types and taking advantage of the language - more on this later) would've been a big task. A Swift rewrite of this piece of the stack was arguably less work.

(Bad) Patterns in the Objective-C codebase were starting to emerge that I knew Swift (by design) would fix. An example of this is the responsibility of view controllers. Some view controllers were table views, some needed [refreshing capabilities](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/TBARefreshTableViewController.h) (but not all), and some had ["container" functionality](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/TBAContainerTableViewController.h) (a pattern for reusing code). In order to get one table view controller that has all 3 things, you'd have to subclass all the way up - 3 classes to accomplish 1 view controller. Additionally we have to write this pattern for all 3 view controller types we were dealing with (UIViewController, UITableViewController, and UICollectionViewController). In Swift, you could just have a Refreshing protocol that you could have these classes conform to (or not) as needed - an obvious win for less code duplication and a significantly simpler project.

Finally, I had to decide if my time was better spent finishing up the Objective-C codebase/releasing a beta/slowly moving to Swift or ditching most of the Objective-C codebase/rewriting in Swift. Migrating a production app is a tough job. Think about our view hierarchy with respect to classes. An example of the way the TBATeamsViewController is setup in the Obj-C branch is TBATeamsViewController subclasses TBARefreshTableViewController, which subclasses TBAContainerTableViewController, which subclasses UITableViewController. The simplest class to rewrite would be the TBAContainerTableViewController, then the TBARefreshTableViewController, then the TBATeamsViewController. Except you can't work that way, since you can't subclass Swift classes in Obj-C. So you have to work suboptimally from the top down (Teams -> Refresh -> Container). While you're working your way down, you can't use Swift-only features that you'd want to be using, like structs. So *really* a rewrite of this example view hierarchy would take two passes. From the top down (Teams -> Refresh -> Container) to get initial functionality, then from the bottom up (Container -> Refresh -> Teams) to get all the Swift goodness. When it comes to view controllers this may not seem important, but moving TBAKit from reference types to value types would be a huge undertaking. Additionally, if I were to release an app and then start to move my Core Data models from APIv2 models to APIv3 models I'd have to write migration methods - which isn't that bad, but it's also someplace we'd be likely to see crashing.

For what it's worth though, most of the architecture decisions made in the Obj-C codebase were still good. A large amount of the Swift codebase is going to be based on work I've done out of the Obj-C codebase, just with a lot less code duplication.

rytonbay:
Thank you for a truly excellent explanation! I like your reasoning/architecture philosophy a lot and agree with your conclusion. It's super cool that APIv3 was written with mobile in mind!

I have some experience with obj-c, and experience with swift. I have experience with relational databases, but the last time I used core data was several years ago. I've worked a bit with TBA APIs in Python before but never made a mobile app with them. I will probably be only contributing slowly for the next couple of months, since I have another project of my own that I need to finish, but I should be able to make some progress each week. Once I'm done with that project I'll be ready to jump more into this one. What would you recommend that I work on/where do you think I would be most useful?

Thank You!
Bryton






On May 11, 2017, at 10:19 PM, Zach Orr <notifications@github.com> wrote:

No worries! These are all good questions.

I'd say the add-google-cloud-platform branch (the most recent branch I was working off of for the Obj-C codebase, which adds MyTBA functionality, but also a lot of other things) is probably 80% of the way "done", but the other 20% is in the details.

There are a lot of reasons for a Swift rewrite... bear with me

The Swift rewrite wasn't a light decision. We'd been hearing from potential contributors that they didn't know Objective-C and were hesitant to get involved. My stance has always been iOS work is mostly knowing the frameworks, but for less experienced programmers picking up unfamiliar syntax can be daunting. Since Swift/Obj-C doesn't matter to me, I weighted all the other factors in making the switch and decided we should go for it.

If you attempt to compile the Objective-C codebase you'll notice a handful of things are broken - but mostly Events. A little bit of code that got added <https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/Models/Event.m#L352> doesn't age super well and would probably get rejected nowadays if I saw it in a PR. Additionally some of the data modeling I was doing <https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/Models/Event.m#L150> in order to sort events was going to have to be rethought, since the initial design was neither eloquent or scalable. This would require rewriting a lot of the Event model, and the view controllers that consume Events.

The biggest influencer for me was APIv3, which we're currently using the native apps to consume/test. APIv3 was designed with mobile in mind, meaning we had a lot of say in how data was structured in order to make our lives easier. Ex: The Team@Event status view in Android was being generated entirely client-side, which is a lot of work. In APIv3 the server will figure this all out for us and return what we need. The Obj-C version of TBAKit <https://github.com/ZachOrr/TBAKit/tree/changes> (the API wrapper I wrote for TBA) was all v2, and upgrading to v3 and moving to Swift (going from reference types to value types and taking advantage of the language - more on this later) would've been a big task. A Swift rewrite of this piece of the stack was arguably less work.

(Bad) Patterns in the Objective-C codebase were starting to emerge that I knew Swift (by design) would fix. An example of this is the responsibility of view controllers. Some view controllers were table views, some needed refreshing capabilities <https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/TBARefreshTableViewController.h> (but not all), and some had "container" functionality <https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/TBAContainerTableViewController.h> (a pattern for reusing code). In order to get one table view controller that has all 3 things, you'd have to subclass all the way up - 3 classes to accomplish 1 view controller. Additionally we have to write this pattern for all 3 view controller types we were dealing with (UIViewController, UITableViewController, and UICollectionViewController). In Swift, you could just have a Refreshing protocol that you could have these classes conform to (or not) as needed - an obvious win for less code duplication and a significantly simpler project.

Finally, I had to decide if my time was better spent finishing up the Objective-C codebase/releasing a beta/slowly moving to Swift or ditching most of the Objective-C codebase/rewriting in Swift. Migrating a production app is a tough job. Think about our view hierarchy with respect to classes. An example of the way the TBATeamsViewController is setup in the Obj-C branch is TBATeamsViewController subclasses TBARefreshTableViewController, which subclasses TBAContainerTableViewController, which subclasses UITableViewController. The simplest class to rewrite would be the TBAContainerTableViewController, then the TBARefreshTableViewController, then the TBATeamsViewController. Except you can't work that way, since you can't subclass Swift classes in Obj-C. So you have to work suboptimally from the top down (Teams -> Refresh -> Container). While you're working your way down, you can't use Swift-only features that you'd want to be using, like structs. So really a rewrite of this example view hierarchy would take two passes. From the top down (Teams -> Refresh -> Container) to get initial functionality, then from the bottom up (Container -> Refresh -> Teams) to get all the Swift goodness. When it comes to view controllers this may not seem important, but moving TBAKit from reference types to value types would be a huge undertaking. Additionally, if I were to release an app and then start to move my Core Data models from APIv2 models to APIv3 models I'd have to write migration methods - which isn't that bad, but it's also someplace we'd be likely to see crashing.

For what it's worth though, most of the architecture decisions made in the Obj-C codebase were still good. A large amount of the Swift codebase is going to be based on work I've done out of the Obj-C codebase, just with a lot less code duplication.

ZachOrr:
Feel free to jump in to our Slack chat - it's a great place to coordinate our efforts https://groups.google.com/forum/#!topic/thebluealliance-developers/-KrhqOSgTcA

rytonbay:
Asked for an invite, will jump on when invited. 

Thanks,
Bryton

PR #81, opened 2017-06-04
Event Rankings

rytonbay:
Here it is with event rankings! Will also submit a PR to TBAKit.


the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -59,9 +59,18 @@
...
+        <relationship name="rankings" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="EventRanking" inverseName="event" inverseEntity="EventRanking" syncable="YES"/>
         <relationship name="teams" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="Team" inverseName="events" inverseEntity="Team" syncable="YES"/>
         <relationship name="webcasts" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="Webcast" inverseName="event" inverseEntity="Webcast" syncable="YES"/>
     </entity>
+    <entity name="EventRanking" representedClassName="EventRanking" syncable="YES" codeGenerationType="class">
+        <attribute name="qualAverage" optional="YES" attributeType="Double" defaultValueString="0" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="rank" optional="YES" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="record" optional="YES" attributeType="Transformable" syncable="YES"/>
```
ZachOrr: Can we handle this `record` transformable the same way we handle transformables for other objects where we get some type safety? Thinking breakdowns in the `Match` object, `divisionKeys` in the `Event` object

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -59,9 +59,18 @@
...
+        <attribute name="tieBreakers" optional="YES" attributeType="Transformable" syncable="YES"/>
```
ZachOrr: Same as above for handling transformables for type safety

the-blue-alliance-ios.xcodeproj/project.pbxproj
```diff
@@ -16,18 +16,19 @@
 		145C57B91ED3C166004955B2 /* MatchInfoViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 145C57B81ED3C166004955B2 /* MatchInfoViewController.swift */; };
 		145C57BB1ED3C178004955B2 /* MatchBreakdownViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 145C57BA1ED3C178004955B2 /* MatchBreakdownViewController.swift */; };
 		145C57BD1ED3C3AA004955B2 /* PlayerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 145C57BC1ED3C3AA004955B2 /* PlayerView.swift */; };
-		149F85861EDF81200012E8BE /* main.jsbundle in Resources */ = {isa = PBXBuildFile; fileRef = 149F857F1EDF81200012E8BE /* main.jsbundle */; };
```
ZachOrr: Oh nooo let's make sure we don't remove the reference to these files please

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,31 @@
...
+import CoreData
+
+extension EventRanking: Managed {
+    
+    static func insert(with model: TBAEventRanking, for event: Event, for team: Team, in context: NSManagedObjectContext) -> EventRanking {
+        let predicate = NSPredicate(format: "event == %@ AND team == %@", event, team)
+        return findOrCreate(in: context, matching: predicate, configure: { (ranking) in
+            ranking.event = event
+            ranking.team = team
+            if let qualAverage = model.qualAverage {
```
ZachOrr: Let's get some spacing up in here! Let this code breathe!

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,31 @@
...
+                ranking.qualAverage = qualAverage
+            }
+            ranking.rank = Int16(model.rank)
+            ranking.record = model.record as NSObject
+            ranking.tieBreakers = model.tieBreakers as NSObject
+
```
ZachOrr: but not these spaces - these spaces at the end of the method can go

the-blue-alliance-ios/View Controllers/Events/EventRankingsTableViewController.swift
```diff
@@ -0,0 +1,165 @@
...
+        
+        clearsSelectionOnViewWillAppear = splitViewController!.isCollapsed
+    }
+    
+    // MARK: - Refreshing
+    
+    override func refresh() {
+        removeNoDataView()
+        
+        // First things first... refresh all teams for the district, *then* fetch their rankings
```
ZachOrr: Thinking about this one... I know we fetch all the teams when fetching district rankings, but I'm wondering if it makes sense here... We might be better off fetching all the teams when we push to the Event VC, and then blocking refreshes for subsequent view controllers (teams, rankings, matches, awards, etc) until we've fetched the teams... for now, let's remove the outer fetch of events and go with the pattern we use in the Awards VC, where when inserting awards we just check if a team exists and if not we make a placeholder team object using the key https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/swift/the-blue-alliance-ios/Core%20Data/Award.swift#L37

ZachOrr: Not that this is a *bad* way to go about fetching relevant data that we need, I've just been rethinking how we actually want to do this

the-blue-alliance-ios.xcodeproj/project.pbxproj
```diff
@@ -714,13 +726,14 @@
 				145C57BB1ED3C178004955B2 /* MatchBreakdownViewController.swift in Sources */,
 				92942DBB1E215B8D008E79CA /* MyTBAViewController.swift in Sources */,
 				924F8C871EC8C71C00F7447A /* NSManagedObjectContext+Extension.swift in Sources */,
-				924F8C8B1EC8D27B00F7447A /* RankingTableViewCell.swift in Sources */,
```
ZachOrr: Do we need two table view cells here... or can we just have the single one? For a District Ranking we can just hide the WLT, and show it for an Event Ranking. I feel like they share enough code that we can make it one table view cell

ZachOrr: The same way the old Obj-C codebase did

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,31 @@
...
+        let predicate = NSPredicate(format: "event == %@ AND team == %@", event, team)
+        return findOrCreate(in: context, matching: predicate, configure: { (ranking) in
+            ranking.event = event
+            ranking.team = team
+            if let qualAverage = model.qualAverage {
+                ranking.qualAverage = qualAverage
+            }
+            ranking.rank = Int16(model.rank)
+            ranking.record = model.record as NSObject
+            ranking.tieBreakers = model.tieBreakers as NSObject
```
ZachOrr: See my comments about type safety stuff regarding these transformables

the-blue-alliance-ios.xcodeproj/project.pbxproj
ZachOrr: But I'll leave this one up to you! If you want to keep them as two seperate files we can. My thought is if we want to update styles for a ranking cell we don't want to update them in two places. Ex: "Let's make the team name 1px smaller"

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -59,9 +59,18 @@
...
         <relationship name="district" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="District" inverseName="events" inverseEntity="District" syncable="YES"/>
         <relationship name="matches" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="Match" inverseName="event" inverseEntity="Match" syncable="YES"/>
+        <relationship name="rankings" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="EventRanking" inverseName="event" inverseEntity="EventRanking" syncable="YES"/>
         <relationship name="teams" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="Team" inverseName="events" inverseEntity="Team" syncable="YES"/>
         <relationship name="webcasts" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="Webcast" inverseName="event" inverseEntity="Webcast" syncable="YES"/>
     </entity>
+    <entity name="EventRanking" representedClassName="EventRanking" syncable="YES" codeGenerationType="class">
+        <attribute name="qualAverage" optional="YES" attributeType="Double" defaultValueString="0" usesScalarValueType="YES" syncable="YES"/>
```
ZachOrr: If this is an optional we probably don't want to use a scalar value - using a NSNumber will allow us to check for it's existence (Core Data doesn't support Swift types yet :| - so even though this is stored as a Double, it will have a default value of 0, not nil)

ZachOrr: We should also remove the "default" if this is an optional - either we have this piece of data or we don't. We don't want Core Data to set the initial value for us

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -59,9 +59,18 @@
...
+        <attribute name="rank" optional="YES" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES" syncable="YES"/>
```
ZachOrr: Let's remove the default for this value too. Also, `rank` shouldn't be optional

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -59,9 +59,18 @@
...
+        <attribute name="record" optional="YES" attributeType="Transformable" syncable="YES"/>
+        <attribute name="tieBreakers" optional="YES" attributeType="Transformable" syncable="YES"/>
+        <relationship name="event" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="Event" inverseName="rankings" inverseEntity="Event" syncable="YES"/>
```
ZachOrr: Let's also remove the optionals from `team` and `event` - a event ranking needs both a team and an event, so we want to throw an error if we don't have one of these pieces when attempting to save

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -112,6 +121,7 @@
         <relationship name="awards" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="AwardRecipient" inverseName="team" inverseEntity="AwardRecipient" syncable="YES"/>
         <relationship name="blueMatches" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="Match" inverseName="blueAlliance" inverseEntity="Match" syncable="YES"/>
         <relationship name="districtRankings" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="DistrictRanking" inverseName="team" inverseEntity="DistrictRanking" syncable="YES"/>
+        <relationship name="eventRankings" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="EventRanking" inverseName="team" inverseEntity="EventRanking" syncable="YES"/>
```
ZachOrr: Make sure this is a To Many - teams can have more than one ranking associated with them (one for each even they've been to)

the-blue-alliance-ios/View Controllers/Events/EventRankingsTableViewController.swift
```diff
@@ -0,0 +1,165 @@
...
+                
+                var rankingsRequest: URLSessionDataTask?
+                rankingsRequest = TBAEvent.fetchRankings(key: self.event.key!, completion: { (rankings, error) in
+                    if let error = error {
+                        self.showErrorAlert(with: "Unable to refresh event rankings - \(error.localizedDescription)")
+                    }
+                    
+                    self.persistentContainer?.performBackgroundTask({ (backgroundContext) in
+                        let backgroundEvent = backgroundContext.object(with: self.event.objectID) as! Event
+                        let realRankings = (rankings?["rankings"] as! [[String: Any]]).map { TBAEventRanking.init(json: $0) }
```
ZachOrr: See my comment on the TBAKit PR, but this seems weird to me https://github.com/ZachOrr/TBAKit/pull/2/files#r120013112

ZachOrr: Also, this will crash in some cases  - you can have events that don't have ranking information, so you're force casting a null to an array https://www.thebluealliance.com/api/v3/event/2004nj/rankings

the-blue-alliance-ios/View Controllers/Events/EventRankingsTableViewController.swift
```diff
@@ -0,0 +1,165 @@
...
+                                backgroundTeam = Team.fetchSingleObject(in: backgroundContext, configure: { (fetchRequest) in
+                                    fetchRequest.predicate = NSPredicate(format: "key == %@" , modelRanking!.teamKey)
+                                })
+                            }
+                            if let backgroundTeam = backgroundTeam {
+                                return EventRanking.insert(with: modelRanking!, for: backgroundEvent, for: backgroundTeam, in: backgroundContext)
+                            }
+                            return nil
+                        })
+                        backgroundEvent.rankings = Set(localRankings ) as NSSet
```
ZachOrr: Remove that random space please :)

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
```diff
@@ -18,18 +18,21 @@ class EventViewController: ContainerViewController {
...
     @IBOutlet internal var teamsView: UIView!
     
-    @IBOutlet internal var rankingsView: UIView?
+    internal var rankingsViewController: EventRankingsTableViewController!
+    @IBOutlet internal var rankingsView: UIView!
     
     internal var matchesViewController: MatchesTableViewController!
     @IBOutlet internal var matchesView: UIView!
     
+
```
ZachOrr: Aaahhh where did these come from?? Can we remove them?

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
```diff
@@ -58,11 +61,12 @@ class EventViewController: ContainerViewController {
             matchesViewController.matchSelected = { match in
                 self.performSegue(withIdentifier: "MatchSegue", sender: match)
             }
-        } else if segue.identifier == "EventMatchesEmbed" {
```
ZachOrr: Let's not remove the Matches embed segue... we need to make sure we still setup the matches VC

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
```diff
@@ -58,11 +61,12 @@ class EventViewController: ContainerViewController {
...
-            matchesViewController = segue.destination as! MatchesTableViewController
-            matchesViewController.event = event
-            matchesViewController.matchSelected = { match in
-                self.performSegue(withIdentifier: "MatchSegue", sender: match)
+        } else if segue.identifier == "EventRankingsEmbed" {
+           rankingsViewController = segue.destination as! EventRankingsTableViewController
+            rankingsViewController.event = event
+            rankingsViewController.rankingSelected = { team in
+                // TODO: Show team@event
+
```
ZachOrr: Remove this space too please

the-blue-alliance-ios/View Controllers/Events/EventsStoryboard.storyboard
```diff
@@ -439,6 +439,24 @@
...
+                        <color key="backgroundColor" white="1" alpha="1" colorSpace="calibratedWhite"/>
+                        <connections>
+                            <outlet property="dataSource" destination="CvT-u0-qAP" id="ncW-7S-VdT"/>
+                            <outlet property="delegate" destination="CvT-u0-qAP" id="qpN-dc-gaT"/>
+                        </connections>
+                    </tableView>
+                </tableViewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="RZI-DR-gwW" userLabel="First Responder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="2134" y="-798"/>
```
ZachOrr: Super picky but... can we move the event ranking table view controller down underneath the event view controller in the storyboard, with the reset of the root view segues? That way they're organized the same.

the-blue-alliance-ios/View Elements/Events/EventRankingTableViewCell.swift
```diff
@@ -0,0 +1,39 @@
...
+        didSet {
+            guard let ranking = ranking, let team = ranking.team else {
+                return
+            }
+            rankLabel?.text = "Rank \(ranking.rank)"
+            numberLabel?.text = "\(team.teamNumber)"
+            nameLabel?.text = team.nickname ?? team.fallbackNickname
+            rankingPointsLabel?.text = "Avg. \(ranking.qualAverage) Points"
+            if let record = ranking.record as? [String: Int] {
+                if let wins = record["wins"], let losses = record["losses"], let ties = record["ties"] {
```
ZachOrr: See my TBAKit comment about adding some type safety around these https://github.com/ZachOrr/TBAKit/pull/2/files#r120012994

the-blue-alliance-ios/View Elements/Events/EventRankingTableViewCell.swift
```diff
@@ -0,0 +1,39 @@
...
+    static let reuseIdentifier = "EventRankingCell"
+    public var ranking: EventRanking? {
+        didSet {
+            guard let ranking = ranking, let team = ranking.team else {
+                return
+            }
+            rankLabel?.text = "Rank \(ranking.rank)"
+            numberLabel?.text = "\(team.teamNumber)"
+            nameLabel?.text = team.nickname ?? team.fallbackNickname
+            rankingPointsLabel?.text = "Avg. \(ranking.qualAverage) Points"
```
ZachOrr: We'll need to check to make sure that qualAverage exists before setting this label up

the-blue-alliance-ios/View Elements/Events/EventRankingTableViewCell.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+
+import Foundation
+import UIKit
+
+class EventRankingTableViewCell: UITableViewCell {
+    static let reuseIdentifier = "EventRankingCell"
+    public var ranking: EventRanking? {
+        didSet {
+            guard let ranking = ranking, let team = ranking.team else {
```
rytonbay: yup

the-blue-alliance-ios/View Elements/Events/EventRankingTableViewCell.swift
rytonbay: You currently it's type is Double, so it I shouldn't have to check right?

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
rytonbay: Interesting, I'm not super familiar with Core Data, and just had to look up what a scalar type really was here. Do you often end up editing the xml like that? It doesn't seem like the kind of development practice that I would normally associate with Apple.

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
rytonbay: yup

rytonbay: Maybe I'm not understanding something, but it seems like this is a different case than something like `divisionKeys`, because division keys doesn't have another model, its just an array of strings, and I'm not exactly sure what breakdowns you're talking about. Should this be done more like the Event's Media?

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
rytonbay: Ah, yes, thanks for catching that.

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
rytonbay: Found the optional checkbox in the editor, so don't have to use xml for that, still being puzzles by the scalar value thing.

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
rytonbay: Makes sense. Done.

the-blue-alliance-ios.xcodeproj/project.pbxproj
rytonbay: must've deleted accidentally when resolving merge conflicts I guess, gosh really didn't mean to. I'll be more careful.

the-blue-alliance-ios.xcodeproj/project.pbxproj
rytonbay: Yeah, I almost made it one file, the reason I didn't is that it would make that one cell a little more complicated, and also that if we want to change one of them but not the other in the future, we won't have to worry about another sometimes-hidden cell feature.

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: Yes!

the-blue-alliance-ios/View Elements/Events/EventRankingTableViewCell.swift
ZachOrr: `qualAverage` should be an optional. It exists for 2015, but not for other years (I believe)

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
ZachOrr: I can explain scalars on Slack if you'd like, or provide some related relating

rytonbay:
Alright, mind taking another look @ZachOrr

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
rytonbay: yes

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
rytonbay: for whatever reason it existed twice it looks like, I was just replacing the duplicate with the rankings segue.

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
rytonbay: done

the-blue-alliance-ios/View Controllers/Events/EventsStoryboard.storyboard
rytonbay: done, totally reasonable Imo

the-blue-alliance-ios/View Controllers/Events/EventViewController.swift
ZachOrr: 👍  That one is probably my fault

the-blue-alliance-ios.xcodeproj/project.pbxproj
```diff
@@ -481,6 +484,7 @@
 		92D855DA1EC6B8F100572F27 /* Event */ = {
 			isa = PBXGroup;
 			children = (
+				303F82491EF593BE007CADA1 /* EventRankingsTableViewController.swift */,
```
ZachOrr: Super picky again but - can we just move this down below the InfoTableViewController in the file hierarchy

the-blue-alliance-ios.xcodeproj/project.pbxproj
```diff
@@ -579,6 +585,7 @@
 				TargetAttributes = {
 					92942D911E2154DA008E79CA = {
 						CreatedOnToolsVersion = 8.2.1;
+						DevelopmentTeam = J3MNNM5PHM;
```
ZachOrr: Can we set this back to None?

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Created by Bryton Moeller on 6/1/17.
+//  Copyright © 2017 The Blue Alliance. All rights reserved.
+//
+
+import Foundation
+import TBAKit
+import CoreData
+
+extension EventRanking: Managed {
+    static func insert(with model: TBAEventRanking, for event: Event, for team: Team, in context: NSManagedObjectContext) -> EventRanking {
```
ZachOrr: Newline before this function declaration, please

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,44 @@
...
+        let predicate = NSPredicate(format: "event == %@ AND team == %@", event, team)
+        return findOrCreate(in: context, matching: predicate, configure: { (ranking) in
+
```
ZachOrr: Remove *this* newline (weird spacing)

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//  Created by Bryton Moeller on 6/1/17.
+//  Copyright © 2017 The Blue Alliance. All rights reserved.
+//
+
+import Foundation
+import TBAKit
+import CoreData
+
+extension EventRanking: Managed {
```
ZachOrr: Can we also set the properties for `dq`, `matchesPlayed`, and `extraStats`?

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,44 @@
...
+            
+            if let ties = model.record?.ties {
+                ranking.ties = ties as NSNumber
+            }
+            
+            if let losses = model.record?.losses {
+                ranking.losses = losses as NSNumber
+            }
+            
+            if let tieBreakders = model.sortOrders as NSObject? {
```
ZachOrr: Typo here - just a misspelling

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -77,6 +78,19 @@
         <relationship name="event" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="Event" inverseName="points" inverseEntity="Event" syncable="YES"/>
         <relationship name="team" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="Team" inverseName="eventPoints" inverseEntity="Team" syncable="YES"/>
     </entity>
+    <entity name="EventRanking" representedClassName="EventRanking" syncable="YES" codeGenerationType="class">
+        <attribute name="dq" optional="YES" attributeType="Integer 16" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="extraStats" optional="YES" attributeType="Transformable" syncable="YES"/>
```
ZachOrr: For these transformables... we can add some type safety around them. Check out what we do in `Event` for `divisionKeys` and `insights` https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/swift/the-blue-alliance-ios/Core%20Data/Event.swift#L27

You have to change the `Custom Class` field in the property editor in Core Data but that's pretty easy
https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/swift/the-blue-alliance-ios/Core%20Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents#L44

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,44 @@
...
+extension EventRanking: Managed {
+    static func insert(with model: TBAEventRanking, for event: Event, for team: Team, in context: NSManagedObjectContext) -> EventRanking {
+        let predicate = NSPredicate(format: "event == %@ AND team == %@", event, team)
+        return findOrCreate(in: context, matching: predicate, configure: { (ranking) in
+            
+            ranking.event = event
+            ranking.team = team
+            
+            if let qualAverage = model.qualAverage {
+                ranking.qualAverage = qualAverage as NSNumber
```
ZachOrr: Let's not cast this from whatever type it is to an NSNumber... let’s instantiate an NSNumber using that value

the-blue-alliance-ios/Core Data/EventRanking.swift
ZachOrr: I'm also skeptical of casting this as an NSObject when it does have an explicit type associated with it - see type safety with transformables comment below

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
```diff
@@ -46,9 +65,18 @@ class RankingTableViewCell: UITableViewCell {
...
+    }
+ 
     @IBOutlet public var rankLabel: UILabel?
     @IBOutlet private var numberLabel: UILabel?
     @IBOutlet private var nameLabel: UILabel?
-    @IBOutlet private var detailLabel: UILabel?
-    @IBOutlet private var detailLabelWidth: NSLayoutConstraint?
+    @IBOutlet private var WLTLabel: UILabel?
+    @IBOutlet var detailLabel: UILabel!
+    @IBOutlet var detailLabelWidth: NSLayoutConstraint!
```
ZachOrr: Can we remove this?

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.xib
```diff
@@ -25,40 +25,52 @@
                                 <rect key="frame" x="0.0" y="0.0" width="68" height="100.5"/>
```
ZachOrr: Can we set the root stack view (the horizontal one) to have an alignment of top? Maybe add like 4px spacing to the other two stack views so the text gets some room to breathe. But that might help some of the spacing weirdness

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
```diff
@@ -46,9 +65,18 @@ class RankingTableViewCell: UITableViewCell {
...
+        }
+        
+    }
+ 
     @IBOutlet public var rankLabel: UILabel?
     @IBOutlet private var numberLabel: UILabel?
     @IBOutlet private var nameLabel: UILabel?
-    @IBOutlet private var detailLabel: UILabel?
-    @IBOutlet private var detailLabelWidth: NSLayoutConstraint?
+    @IBOutlet private var WLTLabel: UILabel?
```
ZachOrr: Can we rename this to be `wltLabel`? Properties don't usually start with uppercase names

the-blue-alliance-ios/View Controllers/Events/EventRankingsTableViewController.swift
```diff
@@ -0,0 +1,144 @@
...
+        super.viewWillAppear(animated)
+        
+        clearsSelectionOnViewWillAppear = splitViewController!.isCollapsed
+    }
+    
+    // MARK: - Refreshing
+    
+    override func refresh() {
+        removeNoDataView()
+        var rankingsRequest: URLSessionDataTask?
```
ZachOrr: Can we put a newline here?

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
```diff
@@ -11,18 +11,36 @@ import UIKit
...
+            guard let ranking = eventRanking, let team = ranking.team else {
+                return
+            }
+            rankLabel?.text = "Rank \(ranking.rank)"
+            numberLabel?.text = "\(team.teamNumber)"
+            nameLabel?.text = team.nickname ?? team.fallbackNickname
+            if let qualAverage = ranking.qualAverage as? Double {
+                detailLabel?.text = "Avg. \(qualAverage) Points"
+            }
+            self.setupWLTLabel(ranking: ranking)
```
ZachOrr: This doesn't need `self`

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
```diff
@@ -46,9 +65,18 @@ class RankingTableViewCell: UITableViewCell {
         }
     }
     
+    func setupWLTLabel(ranking: EventRanking) {
+        if let wins = ranking.wins, let losses = ranking.losses, let ties = ranking.ties {
+            WLTLabel?.text = "(\(wins)-\(losses)-\(ties))"
+        }
+
```
ZachOrr: Remove this newline here

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
```diff
@@ -11,18 +11,36 @@ import UIKit
...
+    public var districtRanking: DistrictRanking? {
+        didSet {
+            guard let ranking = districtRanking, let team = ranking.team else {
                 return
             }
             rankLabel?.text = "Rank \(ranking.rank)"
             numberLabel?.text = "\(team.teamNumber)"
             nameLabel?.text = team.nickname ?? team.fallbackNickname
             detailLabel?.text = "\(ranking.pointTotal) Points"
+
```
ZachOrr: Can we remove this newline?

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
```diff
@@ -11,18 +11,36 @@ import UIKit
...
+            }
+            rankLabel?.text = "Rank \(ranking.rank)"
+            numberLabel?.text = "\(team.teamNumber)"
+            nameLabel?.text = team.nickname ?? team.fallbackNickname
+            if let qualAverage = ranking.qualAverage as? Double {
+                detailLabel?.text = "Avg. \(qualAverage) Points"
+            }
+            self.setupWLTLabel(ranking: ranking)
+        }
+
```
ZachOrr: Can we remove this newline?

the-blue-alliance-ios.xcodeproj/project.pbxproj
rytonbay: Absolutely!

the-blue-alliance-ios.xcodeproj/project.pbxproj
rytonbay: yeah, sorry that would be annoying to find after you pulled

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: Yeah, they were there, but I must've done a bad merge earlier at some point.

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: done

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: done

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: Because Double isn't in the same class hierarchy as NSNumber? What about the Ints?

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
rytonbay: Cool! Thanks for teaching me about this, no its not hard.

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: done.

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
rytonbay: True. It's something I do in my own code for acronyms because otherwise I tend to mentally pronounce them wrong, but I'm happy to change it here.

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
rytonbay: done.

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.xib
rytonbay: tbh I didn't know stack views existed until this project, they're amazing. They're the one thing I thought was missing from iOS ui creation that android has, turns out they weren't missing.

the-blue-alliance-ios/View Controllers/Events/EventRankingsTableViewController.swift
rytonbay: done

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
rytonbay: yup

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
rytonbay: done

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
rytonbay: done. I think I'm starting to get the hang of this project's syntax standards. Sorry it's taken so long.

the-blue-alliance-ios/View Elements/Events/RankingTableViewCell.swift
rytonbay: done

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -0,0 +1,74 @@
...
+            ranking.team = team
+            
+            if let qualAverage = model.qualAverage {
+                ranking.qualAverage = NSNumber(value: qualAverage)
+            }
+            
+            ranking.rank = Int16(model.rank)
+            
+            if let wins = model.record?.wins {
+                ranking.wins = wins as NSNumber
```
ZachOrr: For all of these (`wins`, `ties`, `losses`, `dq`, `matchesPlayed`) - can we create an NSNumber object instead of casting

ZachOrr:
In the Obj-C version I have a method (can probably be a computed property in Swift) called `infoString` on EventRanking that generates a string to fill in the detail label on an event ranking cell. Can we do that here as well? https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/add-google-cloud-platform/the-blue-alliance-ios/Models/EventRanking.m#L119

simulator screen shot may 11 2017 11 15 47 pm

the-blue-alliance-ios/Core Data/EventRanking.swift
rytonbay: done

rytonbay:
@ZachOrr It seems like in order to do this we need access to `sort_order_info` which is not currently part of `EventRanking` or `TBAEventRanking`. We fetch it in parallel with the rankings, and its a `TBAEventRankingSortOrder` (which btw is somewhat confusing because `sortOrders` _is_ within the `EventRanking` object. In the objective-c version there is just one info dictionary that is part of the `EventRanking` object. Should I pass the `TBAEventRankingSortOrder` info object in through the `insert` function? If so, how should I handle type safety?

ZachOrr:
Let's store the `sort_order_info`/`sort_orders` as a dictionary on the event ranking object. We'll just need to pass that sort_order_info object (from TBAKit) to the insert function, and then zip together the two arrays (`sort_order_info`/`sort_orders`) and store them on each ranking object

rytonbay:
So, I pass in sort_order_info to the insert function, and then generate a dictionary where the keys are the "name" properties based of the sort_order_info, and the values are the values in sort_orders. This dictionary exists only as a property in EventRanking, and not in core data. Then the dictionary is used to generate that string that you showed me from the obj-c code. Is that all correct?

rytonbay:
That would mean that we are discarding the "precision" numbers, and discarding the order of the keys. The order seems important since these are tie breakers right?

ZachOrr:
You *would* want to save the `sort_order_info`/`sort_orders` dictionary as a property on `EventRanking` in Core Data. As for precision... let's not worry about it for now. If we need it, it's an easy add

rytonbay:
Dropping precision makes sense to me, but don't we need to retain the order, which we will lose because it's a dictionary?

ZachOrr:
Ummmm true, we will lose order... maybe we store them as two arrays?

rytonbay:
@ZachOrr Is this what you were looking for?

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -30,7 +30,34 @@ extension EventRanking: Managed {
...
+            return sortOrderInfoNamesArray as? Array<String> ?? []
+        }
+        set {
+            sortOrderInfoNamesArray = newValue as NSArray
+        }
+    }
+    
+    var infoString: String? {
+        get {
+            if !tieBreakers.isEmpty && !sortOrderInfoNames.isEmpty && tieBreakers.count == sortOrderInfoNames.count {
```
ZachOrr: Can we use commas here as opposed to `&&` ? It’s more Swift.

the-blue-alliance-ios/Core Data/EventRanking.swift
```diff
@@ -30,7 +30,34 @@ extension EventRanking: Managed {
...
+                var string = ""
+                for (sortOrderName, sortOrderValue) in zip(sortOrderInfoNames, tieBreakers) {
```
ZachOrr: Maybe we try to make these two array names closer to one another so it’s more obvious they belong together? Like, maybe we call `sortOrderInfoNames` something like `tieBreakerNames` and then `tieBreakers` becomes `tieBreakerValues` ? Not sure - I haven’t thought about it much. I’ll leave that one up to you

rytonbay:
Functionality look ok?

Thanks,
Bryton

ZachOrr:
Yes! Looking great. Really great work :)

rytonbay:
@ZachOrr Hopefully we can close this one now :) Thanks for all of your mentorship.

ZachOrr:
This looks great! Go ahead and merge

rytonbay:
Aren't you blocking merge currently?

ZachOrr:
Oops! I think I am... standby...

ZachOrr:
There we go!

rytonbay:
I think you still need to do merge, cuz I don't have write access.

ZachOrr:
Ack okay. Will merge!

ZachOrr:
:shipit:

Issue #128, opened 2018-04-24
To Do Before Ship


ZachOrr:
This will be an updating list with things I need to do as I remember them

- [x] Swap dev `GoogleService-Info.plist` for prod
- [x] Change Bundle Identifier to `com.the-blue-alliance.the-blue-alliance`
- [x] Change React Native Bundle storage bucket fallback to prod
- [x] Add `is_datafeed_down` and `down_events` to realtime database
- [x] Upload prod push certs

ZachOrr:
The big outstanding thing here is the Firebase stuff - we make some assumptions around stuff living in the Realtime Database and Remote Config that may or may not live there in prod.

Chances are we'll have to rewrite these pieces to use the `/status` endpoint

PR #212, opened 2018-06-11
Implement RemoteConfigService, implement AppSetupOperation, rename tests to TestCase, implement remote config in view controllers

### LINKED ISSUE #141: Configure Firebase RemoteConfig
Talked about this in the TBA Slack, but we're looking to move things like minimum iOS versions to Firebase Config... we'll have to keep using the status endpoint for down events and stuff for now

https://github.com/the-blue-alliance/the-blue-alliance-ios/issues/111#issuecomment-385064309

ZachOrr:
Closes #141


ZachOrr:
Thinking about this more... this seems like an appropriate time to kill `Main.storybaord` and add UI tests to make sure everything from the main storyboard gets set up properly in code.

ZachOrr:
I'm also thinking about adding secret management for storing/shipping an API key locally - pulling a key from a server feels not great

https://medium.com/@jules2689/secrets-management-in-ios-applications-52795c254ec1
https://github.com/the-blue-alliance/the-blue-alliance-ios/issues/128

the-blue-alliance-ios/AppDelegate.swift
```diff
@@ -13,105 +13,58 @@ let kNoSelectionNavigationController = "NoSelectionNavigationController"
...
+
+        // Setup our remote config - to be used by our app setup operation
+        let remoteConfigService = RemoteConfigService(remoteConfig: RemoteConfig.remoteConfig(),
+                                                      retryService: RetryService())
+        remoteConfigService.registerRetryable()
+
+        // Our app setup operation will load our persistent stores, fetch our remote config, propogate persistance container
+        let appSetupOperation = AppSetupOperation(window: window,
+                                                  rootSplitViewController: rootSplitViewController(),
+                                                  persistentContainer: NSPersistentContainer(name: "TBA"),
```
ZachOrr: We should think about this one... should we keep a root strong reference to this in the app delegate? It probably doesn't matter since all of the root view controllers have a strong reference but

the-blue-alliance-ios/AppDelegate.swift
```diff
@@ -146,7 +99,60 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     // MARK: Private
 
-    func setupAppearance() {
+    private func rootSplitViewController() -> UISplitViewController {
```
ZachOrr: Let's create this, the tab bar, etc. in code (as opposed to pulling from the storyboard)

the-blue-alliance-ios/AppDelegate.swift
```diff
@@ -146,7 +99,60 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                                                message: "There was an error loading local data - try reinstalling The Blue Alliance",
+                                                preferredStyle: .alert)
+        alertController.addAction(UIAlertAction(title: "Close", style: .default, handler: { (_) in
+            fatalError("Unresolved error \(error), \(error.userInfo)")
+        }))
+        window.rootViewController?.present(alertController, animated: true, completion: nil)
+    }
+
+    private static func setupPushServiceDelegates(with pushService: PushService) {
+        Messaging.messaging().delegate = pushService
```
ZachOrr: We can always move these back up if they're not worth breaking out in to different spots... maybe make it slightly more readable

the-blue-alliance-ios/AppDelegate.swift
```diff
@@ -158,6 +164,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         navigationBarAppearance.titleTextAttributes = [NSAttributedStringKey.foregroundColor: UIColor.white]
     }
 
+    private func launchScreen() -> UIViewController {
```
ZachOrr: We probably don't want to kill the launch screen storyboard - I wonder if we can test this?

the-blue-alliance-ios/AppDelegate.swift
```diff
@@ -262,7 +272,7 @@ extension AppDelegate: UISplitViewControllerDelegate {
         return emptyDetailViewController()
     }
 
-    public func emptyDetailViewController() -> UIViewController {
+    private func emptyDetailViewController() -> UIViewController {
         return UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: kNoSelectionNavigationController)
```
ZachOrr: We can break this out in to an xib when killing `Main.storyboard`

the-blue-alliance-ios/Operations/App Setup/AppSetupOperation.swift
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+    }
+
+    override func execute() {
+        let blockOperation = BlockOperation {
+            if self.completionError == nil {
+                self.propogatePersistentContainer(to: self.rootSplitViewController)
+
+                DispatchQueue.main.async {
```
ZachOrr: Let's move ALL of this back in to the App Delegate - it feels very strange poking around in UIKit stuff inside of here. It also makes it extremely difficult to test

the-blue-alliance-ios/Services/PushService.swift
```diff
@@ -61,6 +56,8 @@ class PushService: NSObject {
                     Crashlytics.sharedInstance().recordError(error)
                     self?.registerRetryable()
                 } else {
+                    // TODO: Make sure this timer gets removed from memory...
```
ZachOrr: Let's write a test for this or something - even if this is the only thing we test right now for the push service. Or, hell, do it manually. But confirm this before merging.

the-blue-alliance-ios/Services/RemoteConfigService.swift
```diff
@@ -0,0 +1,48 @@
...
+        if let configSettings = RemoteConfigSettings(developerModeEnabled: true) {
+            self.remoteConfig.configSettings = configSettings
+        }
+        #endif
+        self.retryService = retryService
+
+        remoteConfig.setDefaults(fromPlist: "RemoteConfigDefaults")
+    }
+
+    internal func fetchRemoteConfig(completion: ((_ error: Error?) -> Void)? = nil) {
```
ZachOrr: Make sure we test this zz

the-blue-alliance-ios/Services/RemoteConfigService.swift
```diff
@@ -0,0 +1,48 @@
...
+
+            if let completion = completion {
+                completion(error)
+            }
+        }
+    }
+
+}
+
+extension RemoteConfigService: Retryable {
```
ZachOrr: I'm not entirely sure if these things are worth adding in to the tests, but if we can we might as well

the-blue-alliance-ios/View Controllers/Districts/DistrictsContainerViewController.swift
```diff
@@ -1,14 +1,14 @@
...
+import FirebaseRemoteConfig
 
 private let SelectYearSegue = "SelectYearSegue"
 
 class DistrictsContainerViewController: ContainerViewController {
 
-    // TODO: Get these from Firebase Config
-    var maxYear: Int = 2018
-    var year: Int = 2018 {
+    var maxYear: Int = RemoteConfig.remoteConfig().maxSeason
```
ZachOrr: Eventually this singleton access will go away, but that won't be until we start doing constructor injection, which we need constructors for, which we need to remove storyboards for

the-blue-alliance-ios/Services/PushService.swift
ZachOrr: I can't see that the timer gets removed from memory, but it seems like our retry service isn't holding on to it, and I can't find a reference in our memory graph where the timer is holding on to the push service soooo

ZachOrr: Yeah - it looks like the block referencing the push service gets properly deallocated when we invalidate the timer and whatnot. I'm not going to worry about why the timer is still in memory - that's the run loops problem at that point

the-blue-alliance-ios/Services/RemoteConfigService.swift
ZachOrr: Probably not going to test this - RemoteConfig isn't really mockable because it's `init` is marked as unavailable

the-blue-alliance-ios/Services/RemoteConfigService.swift
ZachOrr: Did it anyways (kinda)

PR #214, opened 2018-07-18
Add myTBA unregister as syncronous action

### LINKED ISSUE #175: Implement myTBA unregister
This might need some web work - see the unregister comments in PushService for the current thoughts around this. We might look to make that unregister endpoint an unauthenticated endpoint

ZachOrr:
Making unregister a synchronous action via an action and an activity spinner, since signing out of Google makes it where myTBA is unauthenticated, and we need to be authenticated when we hit that unregister endpoint.  Closes #175

### DIFF
--- a/the-blue-alliance-ios/Frameworks/MyTBAKit/Models/MyTBARegister.swift
+++ b/the-blue-alliance-ios/Frameworks/MyTBAKit/Models/MyTBARegister.swift
@@ -12,18 +12,20 @@ struct MyTBARegisterRequest: Codable {
 
 struct MyTBARegisterResponse: MyTBAResponse, Codable {
 
-    var code: Int
+    var code: String
     var message: String
 
 }
 
 extension MyTBA {
 
+    @discardableResult
     func register(_ token: String, completion: @escaping (_ error: Error?) -> Void) -> URLSessionDataTask? {
         return registerUnregister("register", token: token, completion: completion)
     }
 
-    func unregister(_ token: String, completion: @escaping (_ error: Error?) -> Void) -> URLSessionDataTask? {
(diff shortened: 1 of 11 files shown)


the-blue-alliance-ios/Frameworks/MyTBAKit/Models/MyTBARegister.swift
```diff
@@ -12,18 +12,20 @@ struct MyTBARegisterRequest: Codable {
...
 extension MyTBA {
 
+    @discardableResult
     func register(_ token: String, completion: @escaping (_ error: Error?) -> Void) -> URLSessionDataTask? {
         return registerUnregister("register", token: token, completion: completion)
     }
 
-    func unregister(_ token: String, completion: @escaping (_ error: Error?) -> Void) -> URLSessionDataTask? {
+    @discardableResult
+    @objc func unregister(_ token: String, completion: @escaping (_ error: Error?) -> Void) -> URLSessionDataTask? {
```
ZachOrr: the @objc is so that we can override this method in our mock, since you can't override a method in an extension in Swift. The real answer here is probably to not mock the myTBA stuff

the-blue-alliance-ios/Frameworks/MyTBAKit/Models/MyTBARegister.swift
```diff
@@ -12,18 +12,20 @@ struct MyTBARegisterRequest: Codable {
 
 struct MyTBARegisterResponse: MyTBAResponse, Codable {
 
-    var code: Int
+    var code: String
```
ZachOrr: This was also coming back as an error - very strange

Issue #67, opened 2016-05-05
Move "Week 0" to "Offseason Events" section in Preseason week


ZachOrr:
The Android app does this, and it makes sense
![screenshot_20160504-201023](https://cloud.githubusercontent.com/assets/516458/15032972/50bffa9e-1234-11e6-913e-f6f67b03bbdf.png)

Issue #96, opened 2018-04-19
Quarterfinal match names being split to two lines


ZachOrr:
This is on a 7+ - let's either use dynamic type or something to make sure this doesn't happen on any devices

![simulator screen shot - iphone 7 plus - 2018-04-19 at 18 09 41](https://user-images.githubusercontent.com/516458/39021799-733a8efe-4400-11e8-98f9-322efe630b8b.png)

ZachOrr:
Let's try to convert this cell to dynamic type while we're at it

Issue #95, opened 2018-04-19
Video player button not showing up for matches


ZachOrr:
Matches with videos should have a video player icon on the left hand side of the cell. I believe there's an image view that's already there. I'm seeing console information about not referencing the image properly. Should be an easy fix.

PR #215, opened 2018-07-28
Fixed issues #67, #95, and #96

SentientBacon:
- (#67) Changed the way `EventsTableViewController` presents the event list in the table view by adding table view section titles for Offseason Events, Preseason Events, and Festival of Champions.
- (#95) Added `ic_action_play_over_video` assets, allowing the play button to show on events with available videos.
- (#96) Match name labels in `MatchTableViewCell` now force a single line; instead of wrapping, the text now autoshrinks and compresses letter spacing


ZachOrr:
Acknowledging I saw this - I’m flying today, so review might take a moment. I wanna be able to build and run and check the changes out.

Thanks for the PR!

the-blue-alliance-ios/View Controllers/Events/EventsTableViewController.swift
```diff
@@ -218,7 +218,7 @@ extension EventsTableViewController: TableViewDataSourceDelegate {
         guard let event = dataSource?.object(at: IndexPath(item: 0, section: section)) else {
             return nil
         }
-
+
```
ZachOrr: Can we remove bonus space like the? I think there’s a setting in Xcode that will keep this from happening. I haven’t been able to setup a pre-commit linter yet

the-blue-alliance-ios/Core Data/Match.swift
```diff
@@ -53,15 +53,15 @@ extension Match: Managed {
         }
         switch compLevel {
         case MatchCompLevel.qualification.rawValue:
-            return "Qual"
+            return "Quals"
         case MatchCompLevel.eightfinal.rawValue:
-            return "Octofinal"
```
ZachOrr: We want these strings to match Android/web, so this might roll back. Adding a pluralization seems right to me... probably something I’ll check out when I build

ZachOrr: Oh - the web changed this 😅 Yeah let's go ahead and keep this then

ZachOrr:
Thank you! I went ahead and merged for you, but moving forward, feel free to merge after you have an approval ✨

SentientBacon:
Thanks! I would have merged, but I don’t actually have write access.

ZachOrr:
Ah okay! Sorry - I forgot what GH settings are for stuff

PR #187, opened 2018-05-20
Fixed issues #43, #95, #119 & #123

anas-merbouh:
Fixed a bug where table view cell wouldn't auto-resize as well. 
Library : [https://github.com/aFrogleap/SimpleImageViewer](url)

Issue #123, opened 2018-04-22
Add "Delete network cache" button


ZachOrr:
To do in some settings controller - allow the network cache to be deleted (I see some API errors sometimes with this - not sure how much of this is dev)

PR #223, opened 2018-09-20
Adds Settings view controller - closes #119, #123

### LINKED ISSUE #119: Add "Delete Data" button
Drop the existing Core Data DB and start a fresh one - should be good for debugging, but also if someone's device gets in to a bad state, there's something to try

ZachOrr:
Also closing out #187, since it's outdated


the-blue-alliance-ios/Core Data/NSPersistentContainer+Extension.swift
```diff
@@ -0,0 +1,23 @@
+import CoreData
+import Foundation
+
+extension NSPersistentContainer {
```
ZachOrr: These need tests :)

the-blue-alliance-ios/View Controllers/Settings/SettingsViewController.swift
```diff
@@ -0,0 +1,191 @@
...
+            urlOpener.open(url, options: [:], completionHandler: nil)
+        }
+    }
+
+    internal func deleteNetworkCache() {
+        TBAKit.clearLastModified()
+    }
+
+    internal func deleteAppData() {
+        persistentContainer.deleteAllEntities()
```
ZachOrr: If the `NSPersistentContainer` extension gets tests, we can mock this out and just make sure it gets called for this method. Should be easy.

ZachOrr:
We need to think about what "deleting all data" means safely. We might have to look to do something like blow up the underlying persistent store. A lot of our Core Data models aren't safe to delete one by one (singular Entities at a time) because of inverse relationships between models

ZachOrr:
After deleting data and then doing a refresh, the "Events" tab week selector defaults to "Other Events". This needs to get fixed before merging.

ZachOrr:
The "Other Events" issue is actually hanging on to a nil object in the events container view controller. This is a bit of a larger problem... anywhere that hangs on to a Core Data object needs to register for deletion changes.

Or, we watch for the delete event and post notifications to the navigation controllers and pop them all down to the root level, and do the necessary updates to the events container view controller.

Still thinking about the best way to handle this one.

ZachOrr:
Going to get this one mergeable without the delete data stuff - this has a lot of important changes (adding setup for UI tests, reworking the Core Data test case, URLOpener for other storyboard removal work) that I want to get in

PR #227, opened 2018-09-22
Update to new the-blue-alliance-react, fix ReactNativeService to correctly identify downloaded bundle

ZachOrr:
That ReactNativeService sure does need some tests


ZachOrr:
I'm not particularly worried about keeping the RN stuff too up to date - using an insecure version of a js library on the web is much more concerning than using an insecure version of a js library compiled down on a phone with no external user input.

PR #228, opened 2018-09-22
Fix storyboards to use safe area layout guides properly

ZachOrr:
No more white space down by the home bar on the X


ZachOrr:
These need snapshot tests or something, but... soon

Issue #172, opened 2018-05-08
Aduit performSegue calls


ZachOrr:
We should probably be calling `shouldPerformSegue` before calling `performSegue`, since a call to latter jumps over the check on the former

PR #233, opened 2018-09-30
Kills storyboards - closes #209

### LINKED ISSUE #92: Build unified Year -> Week selector
We don't want to break the existing year selector (used for multiple screens, like Team, Districts, etc.), but build a unified Year -> Week selector for the Events VC. Having them be two different buttons is confusing.

I envision when you click the selector, we'll bring you to an already-pushed week selector. If you click Back, you'll get to Years. Clicking a Year from the Years TVC will bring you to the Weeks selector. Selecting a week should dismiss the controller.

### LINKED ISSUE #112: Pass `persistentContainer` automatically
It would be great to pass `persistentContainer` automatically between container view controllers. The Obj-C version does this, but not the Swift version. I have to imagine it's possible. We shouldn't be responsible for setting that in the segues every time

### LINKED ISSUE #195: Rename container/child controllers
It's difficult right now to tell what's a container and what's a child controller. Find some naming convention to make it easy to tell which is which

### LINKED ISSUE #209: Eventually remove storyboards
Discussed with @ZachOrr on Slack. Would be nice to have for testability ! **(would be a post-prod goal)**

ZachOrr:
- Kills all storyboards
- Refactor AppDelegate to setup view hierarchy without storyboard
- Setup injected properties in AppDelegate, to be overriden for testing
- Adds functionality to sort Offseason events in to months
- Moves `weekEvents` to be a static method on Event
- Adds `month` property on Event to get month string for Event
- Reverts `saveContext` to be `saveOrRollback`
- Adds `startOfMonth` and `endOfMonth` in Date extension, in order to fetch offseason events by month
- Removes force unwrapping of `persistentContainer` in Persistable
- Removes force unwrapping of `collectionView` in collection view controllers
- Removes `dataView` from Stateful protocol because it’s unnecessary
- Removes `reactBridge` from ReactNative protocol, since new React Native doesn’t allow us to use things like a fallback URL anymore, and we never needed it
- Makes some enums private to their own files
- Refactors ContainerViewController to be used without a storyboard
- Refactors view controllers to use delegation instead of blocks for callbacks
- Removes `cellIdentifier` from TableViewDataSource and CollectionViewDataSource
- Modifies TBAViewController, TBATableViewController, and TBACollectionViewController for no storyboards and cell refactors
- Refactors SelectNumberTableViewController
- Splits EventsViewController in to DistrictEventsViewController, YearEventsViewController, and TeamEventsViewController to make those classes more sensible
- Creates new system for reusing cells which uses view models for state and a simpler way to register reusable (see Reusable protocol)
- Creates YearSelectViewController for combined year + week selection for events (closes #92)
- Renames view controllers - for the most part, closes #195

Additionally closes #112, #172


ZachOrr:
The MediaCollectionViewController's flow layout isn't getting setup properly, but I'm going to go ahead and merge anyway to get this long running branch out

PR #240, opened 2018-10-05
Initially refresh view controllers (part of #197)

### LINKED ISSUE #197: Use heuristics to no data reload view controllers
We can have a pretty good idea of when to reload a view controller based on heuristics. For instance, if we navigate to the Teams view controller, and we have less than... 500 teams, we're probably working with an incomplete data set. Additionally, if we navigate to an event's teams view controller and we have less than... 30 teams, we're probably working with an incomplete data set. The latter is year-dependent, which is worth thinking about when implementing these.

ZachOrr:
- Initially refresh view controllers (part of #197)
- Removes table view/collection view from `TableViewDataSource`/`CollectionViewDataSource`
- Upgrade TBAKit to 2.0.7
- Conform error enums to `LocalizedError` (and adds tests)
- Renames `APIError` in MyTBA to `MyTBAError`
- Explicitly conform data controllers to `Refreshable`
- Removes `shouldNoDataRefresh` in `Refreshable`, use `isDataSourceEmpty` instead
- Adds `Refreshable where` extension in `TBAViewController`/`TBATableViewController`/`TBACollectionViewController`

### DIFF
--- a/the-blue-alliance-ios/Protocols/Refreshable.swift
+++ b/the-blue-alliance-ios/Protocols/Refreshable.swift
@@ -111,7 +111,6 @@ extension Refreshable {
 
     func enableRefreshing() {
         let refreshControl = UIRefreshControl()
-        // TODO: THIS IS STILL FUCKING BROKEN
         refreshControl.addTarget(self, action: Selector(("refresh")), for: .valueChanged)
 
         self.refreshControl = refreshControl
--- a/the-blue-alliance-ios/Protocols/Refreshable.swift
+++ b/the-blue-alliance-ios/Protocols/Refreshable.swift
@@ -12,11 +12,17 @@ protocol Refreshable: AnyObject {
     var refreshControl: UIRefreshControl? { get set }
     var refreshView: UIScrollView { get }
 
-    func refresh()
-    func shouldNoDataRefresh() -> Bool
+    /**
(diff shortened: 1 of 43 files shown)


the-blue-alliance-ios/View Controllers/Events/YearSelectViewController.swift
```diff
@@ -141,11 +141,22 @@ extension EventWeekSelectViewController: SelectTableViewControllerDelegate {
...
+        return weeks.isEmpty
+    }
+
     func refresh() {
         var request: URLSessionDataTask?
         request = TBAKit.sharedKit.fetchEvents(year: year, completion: { (events, error) in
             if let error = error {
                 self.showErrorAlert(with: "Unable to refresh events - \(error.localizedDescription)")
+            } else {
+                // TODO: Fix this
```
ZachOrr: - [ ] Fix this

the-blue-alliance-ios/Protocols/Refreshable.swift
```diff
@@ -90,4 +109,16 @@ extension Refreshable {
         }
     }
 
+    func enableRefreshing() {
+        let refreshControl = UIRefreshControl()
+        // TODO: THIS IS STILL FUCKING BROKEN
```
ZachOrr: - [ ] Fix pull to refresh

PR #241, opened 2018-10-06
Change refresh pattern to refresh automatically and aggressively

### LINKED ISSUE #197: Use heuristics to no data reload view controllers

ZachOrr:
- Adds more sophisticated way to refresh views using a refresh interval and an end date for automatic refreshes (Closes #197)
- Changes `Event.isHappeningNow` functionality to include `endDate` in calculation
- Adds `Calendar.kickoff` and `Calendar.stopBuildDay` to help figure out when Districts are “happening”
- Adds tests for `Date.isBetween` working inclusively
- Adds `District.endDate` (the end date of the DCMP)


ZachOrr:
FWIW - there's an awful lot of unnecessary loading you see when clicking through the app. One example is District Rankings -> Team@District, we end up refreshing both sub-views, even though we've already fetched that data in the District Rankings controller. That's kind of what the idea in #197 was with keying based on TBAKit refreshes - data for view controllers isn't necessarily tied to the view (we may fetch more than we need, and only show a little bit in a view).

I'm going to merge this one anyways, because it's better than we had before, and it gets us closer to a goal. But I'll keep thinking about keying based on something smarter.

PR #253, opened 2018-10-10
Fix model relationships being overwritten by empty cache responses

ZachOrr:
Previously in a lot of places we were setting up model relationships twice. Once in the `insert` method, and then once in the TBAKit callback. This is problematic when we a 304 from the API, since we will get back an empty array. Our TBAKit callback won't insert any objects, but we'll overwrite the model's relationship. An example of this is in the District Events view

`backgroundDistrict.events = Set(localEvents ?? []) as NSSet`

If we didn't get a modified response from the API, `localEvents` will be nil (or empty), and we'll overwrite all the previous relationships for Events in that district.

This makes our `insert` model the source of truth for setting up most relationships. Other places where we're not passing a Core Data object on inserts we're doing append-only for setting up relationships.

### DIFF
--- a/the-blue-alliance-ios/Core Data/Award.swift
+++ b/the-blue-alliance-ios/Core Data/Award.swift
@@ -4,6 +4,7 @@ import TBAKit
 
 extension Award: Managed {
 
+    @discardableResult
     static func insert(with model: TBAAward, for event: Event, in context: NSManagedObjectContext) -> Award {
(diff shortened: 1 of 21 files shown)


ZachOrr:
Note to Zach: This is PROBLEMATIC. What we really need to do is see if we have data that gets passed back and update our relationships accordingly. aka, this is wrong, what we had was wrong, and the fix is somewhere inbetween

Issue #118, opened 2018-04-21
Troubleshoot Team Media


ZachOrr:
Team Media seems to be having some weird issues with 1) threading (a lot of custom UI stuff not being run on the main thread) and 2) getting data properly. Show's no data when we should have data 

![simulator screen shot - iphone 8 plus - 2018-04-21 at 17 22 30](https://user-images.githubusercontent.com/516458/39088891-93bc0d9a-4588-11e8-953f-df1141a8b384.png)

ZachOrr:
Closing this one out feels good :v:

PR #256, opened 2018-10-12
Fix TeamMediaCollectionViewController (#118)

ZachOrr:
- Fixes `TeamMediaCollectionViewController` load/refresh pattern
- Convert player/media view to player/media collection view cells
- Fix broken build on master
- Fix bug where collection view controller would never show no data

The player/media view changes just make sense, now that we have that `Reusable` protocol.

The no data bug in `UICollectionViewController` is because `collectionView.backgroundColor` sits in front of `collectionView.backgroundView` in the view hierarchy, unlike `UITableViewController`. Changing the background colors to clear for `TBA{View/TableView/CollectionView}Controller` makes these classes unsafe to use when not contained in a `ContainerViewController`. There's probably a better fix for this.

Still to do
- [x] Confirm and find alternative for no data background colors
- [x] Store UIImages to Media objects (get rid of that NSCache) (look in to how to do this efficiently)
- [x] Remove hanging around TODO comments
- [ ] Support YouTube videos in `TeamMediaCollectionViewController`
- [ ] Refactor `PlayerCollectionViewCell`
- [ ] Support multiple cells in `CollectionViewDataSource`, or don't use `CollectionViewDataSource` in `TeamMediaCollectionViewController`

YouTube video stuff might get it's own Issues, and can be handled later.


ZachOrr:
Reverting the no data background changes - I forgot to file a bug for this, but our Stateful wasn't matching that UICollectionView conditional. The view we were trying to match was a container view.

I refactored out the Stateful stuff, the same way I refactored out the Refreshable stuff. Seems like a better pattern (although large, for this PR)

ZachOrr:
Filed a ticket to add YouTube videos - let's go without them, for now

PR #293, opened 2018-11-20
Setup beta builds via Travis

ZachOrr:
- [x] Setup credentials in Travis
- [x] Format changelog details
- [x] Confirm GoogleService-Info.plist
- [x] Fix "Could not find build" from pilot (https://github.com/fastlane/fastlane/issues/13688)
- [x] Populate "What to Test"


fastlane/Fastfile
```diff
@@ -46,17 +38,106 @@ platform :ios do
...
 
+  desc "Create a new app version (major, minor, patch, or anything else for a version bump)"
+  lane :new_version do |options|
+    version_type = options[:version_type]
+    if version_type != "build"
+      increment_version_number(bump_type: version_type)
+    end
+    increment_build_number
+    commit_version_bump
+    push_to_git_remote
```
ZachOrr: This needs to skip CI

fastlane/Fastfile
```diff
@@ -46,17 +38,107 @@ platform :ios do
...
```
ZachOrr: This needs to be moved out of here - it's going to end up being a problem with that `while` loop below. We'll end up committing until we find a proper version which... isn't great. We should probably only commit if we had to bump.

PR #301, opened 2018-11-22
Add methods to fetch and insert bulk MyTBA objects

ZachOrr:
- [x] https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/301#issuecomment-441098987

PR #303, opened 2018-11-22
Better view controller snapshot testing

ZachOrr:
Very much WIP, but let's try to land better VC testing with the new Core Data testing stuff in #301 so we can better test our view controllers


tba-unit-tests/TBAViewControllerTester.swift
```diff
@@ -0,0 +1,54 @@
...
+    private(set) var window: UIWindow!
+
+    init(withViewController rootViewController: T) {
+        window = UIWindow(frame: UIScreen.main.bounds)
+        window.isHidden = false
+        window.rootViewController = rootViewController
+
+        if rootViewController is UINavigationController {
+            let navigationController = rootViewController as! UINavigationController
+            appearViewLifecyle(navigationController)
```
ZachOrr: Do we need to call this, or will the VC call this for us?

ZachOrr: It will not - although, I'm not sure how important driving that navigation controller lifecycle is

tba-unit-tests/TBAViewControllerTester.swift
```diff
@@ -0,0 +1,54 @@
...
+
+        window.rootViewController = nil
+        window.isHidden = true
+        self.window = nil
+    }
+
+    private func appearViewLifecyle(_ viewController: UIViewController) {
+        _ = viewController.view
+        viewController.viewWillAppear(false)
+        viewController.viewDidAppear(false)
```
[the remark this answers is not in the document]
ZachOrr: This just appears to be a problem with when we're setting up that title view. If we setup that `navigationItem` in the init method, the `titleView` gets setup on our first redraw, as opposed to waiting until the first redraw after `viewDidLoad`. This feels like a better fix for the slow title loading then we had before.

tba-unit-tests/TBAViewControllerTester.swift
ZachOrr: Ah - it turns out our window will drive these for the root vc. We'll change the order of these so we're explicitly driving our lifecycle methods when testing

ZachOrr:
This one is good to go 👍

PR #305, opened 2018-11-23
Adds some Core Data testing methods and classes

ZachOrr:
Removing just the Core Data code from #301 to make it smaller and easier to review


ZachOrr:
This looks good 👍

PR #329, opened 2018-11-29
Fix Match view controller crashes assuming Matches have Events

ZachOrr:
Apparently we didn't hit all of the references assuming a Match has an Event attached to it - this might fix some of our flaky tests as well


ZachOrr:
Failing due to flaky tests still - will look in to it later, but merging this for now

PR #331, opened 2018-11-29
Move styleInterface calls to viewDidLoad

ZachOrr:


ZachOrr:
Flaky tests - merging

PR #333, opened 2018-11-29
Fix bug in Core Data observer dispatching all observation blocks, not just the ones for the observed state

ZachOrr:


ZachOrr:
This is important everywhere, but it'll be important for handling deletions of Matches in those match info views

Issue #341, opened 2018-11-30
Change "MyTBA" to "myTBA"


ZachOrr:
We should at least be consistent with branding

Issue #340, opened 2018-11-30
Remove "About Google Accounts" copy from myTBA screen


ZachOrr:
per a request in Slack

PR #343, opened 2018-11-30
Fixes myTBA copy

ZachOrr:


ZachOrr:
Closes #340 and closes #341

PR #351, opened 2018-12-02
Register myTBA after Google sign in, fix myTBA Team@Event crash

### LINKED ISSUE #349: Fix Team@Event myTBA seralization
We're seeing crashes in myTBA related to users fetching Team@Event in the app. We should get a fix for this so that we don't crash. Related to #219.

We might want to be a little bit more relaxed when de-serializing myTBA data, since I can see situations where we add stuff that we don't support.

ZachOrr:
Closes #349

### DIFF
--- a/the-blue-alliance-ios/View Controllers/MyTBA/MyTBATableViewController.swift
+++ b/the-blue-alliance-ios/View Controllers/MyTBA/MyTBATableViewController.swift
@@ -67,6 +67,11 @@ class MyTBATableViewController<T: MyTBAEntity & MyTBAManaged, J: MyTBAModel>: TB
             NSSortDescriptor(key: #keyPath(MyTBAEntity.modelKey), ascending: true)
         ]
 
+        // Only show supported myTBA entities (basically, exclude team@event)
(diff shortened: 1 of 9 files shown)


the-blue-alliance-ios/View Controllers/MyTBA/MyTBATableViewController.swift
```diff
@@ -67,6 +67,11 @@ class MyTBATableViewController<T: MyTBAEntity & MyTBAManaged, J: MyTBAModel>: TB
             NSSortDescriptor(key: #keyPath(MyTBAEntity.modelKey), ascending: true)
         ]
 
+        // Only show supported myTBA entities (basically, exclude team@event)
+        fetchRequest.predicate = NSPredicate(format: "%K IN %@",
```
ZachOrr: We should probably add some test around this...

ZachOrr: Note to Zach - we're adding a test to make sure a Team@Event doesn't show up in snapshot tests. All of the FRC stuff is private, which it feels wrong making it not-private, so we'll test it this way

PR #393, opened 2018-12-16
Adds custom configs to disable analytics in dev

ZachOrr:


ZachOrr:
fwiw simulator logs show this is failing because

```
Terminating app due to uncaught exception 'FABException', reason: '[Fabric] Value of Info.plist key "Fabric" must be a NSDictionary.'
```

I'm not entirely sure which of these changes (the xcconfig or the Info.plist additions) are causing this to break. Fabric should be able to figure out what it needs from Firebase

PR #399, opened 2018-12-19
Adds dynamic capability for adding an alternate icon

ZachOrr:


the-blue-alliance-ios/Info.plist
```diff
@@ -8,6 +8,28 @@
 	<string>TBA</string>
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIcons</key>
+	<dict>
+		<key>CFBundleAlternateIcons</key>
+		<dict>
+			<key>DESTINATION: DEEP SPACE</key>
+			<dict>
+				<key>CFBundleIconFiles</key>
```
ZachOrr: Why can't we use CFBundleIconName here? Would it be worth putting these in the asset catalog? Seems like it

PR #403, opened 2018-12-30
Add RP dots to score listings

### LINKED ISSUE #126: Add Extra RP Dots
I really like this idea - it shouldn't be too hard to implement on iOS

https://github.com/the-blue-alliance/the-blue-alliance-android/pull/866

dracco1993:
This PR adds in RP dots to the score listings.

There was some concern in the original issue ( #126 ) about parsing the JSON on-the-fly, so that's something we might want to at least look at on an actual device.  Locally, on my emulator, it doesn't appear to have changed the render time, but I feel like that's hard to measure.

This closes #126

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ tba-unit-tests/FailureDiffs
 # Ignore some Metadata from Fastlane with sensitive information
 fastlane/metadata/trade_representative_contact_information
 fastlane/metadata/review_information
+
+# rbenv
+.ruby-version
--- a/the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
+++ b/the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
@@ -79,7 +79,7 @@ class MatchSummaryView: UIView {
     private func removeTeams() {
         for stackView in [redStackView, blueStackView] as [UIStackView] {
             for view in stackView.arrangedSubviews {
-                if [redScoreLabel, blueScoreLabel].contains(view) {
+                if [redScoreLabel, redRPLabel, blueScoreLabel, blueRPLabel].contains(view) {
                     continue
                 }
                 view.removeFromSuperview()
--- a/the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
+++ b/the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
@@ -35,8 +35,13 @@ class MatchSummaryView: UIView {
         }
     }
     @IBOutlet weak var blueScoreLabel: UILabel!
+    
+    @IBOutlet weak var redRPLabel: UILabel!
+    
+    @IBOutlet weak var blueRPLabel: UILabel!
 
     @IBOutlet weak var timeLabel: UILabel!
+    
 
     // MARK: - Init
 
@@ -97,12 +102,14 @@ class MatchSummaryView: UIView {
             redStackView.insertArrangedSubview(teamLabel, at: 0)
         }
         redScoreLabel.text = viewModel.redScore
+        redRPLabel.text = viewModel.redRP
 
         for teamKey in viewModel.blueAlliance {
             let teamLabel = label(for: teamKey, baseTeamKey: viewModel.baseTeamKey)
             blueStackView.insertArrangedSubview(teamLabel, at: 0)
         }
         blueScoreLabel.text = viewModel.blueScore
+        blueRPLabel.text = viewModel.blueRP
 
         timeLabel.isHidden = viewModel.hasScores
         timeLabel.text = viewModel.timeString
--- a/the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
+++ b/the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
@@ -27,6 +27,7 @@ class MatchSummaryView: UIView {
         }
     }
     @IBOutlet weak var redScoreLabel: UILabel!
+    @IBOutlet private weak var redRPStackView: UIStackView!
 
     @IBOutlet weak private var blueStackView: UIStackView!
     @IBOutlet weak var blueContainerView: UIView! {
@@ -35,6 +36,7 @@ class MatchSummaryView: UIView {
         }
     }
     @IBOutlet weak var blueScoreLabel: UILabel!
+    @IBOutlet private weak var blueRPStackView: UIStackView!
     
     @IBOutlet weak var redRPLabel: UILabel!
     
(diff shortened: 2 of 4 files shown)


the-blue-alliance-ios/View Elements/Match/MatchViewModel.swift
```diff
@@ -44,6 +47,29 @@ struct MatchViewModel {
...
             return true
         }()
+        
+        // Set RP things, specifically for 2018
+        // Other years can use a similar pattern
+        if match.year == 2018 {
+            redRPString = {
+                guard var redBreakdown = match.breakdown?["red"] as? [String: Any] else { return "" }
+                var rpString: String = ""
+                rpString += (redBreakdown["autoQuestRankingPoint"] as! Bool ? "•" : "")
```
ZachOrr: This "•" string specifically can live in the match cell. There's no reason for us to be passing it around - we can just say "add an RP label to this view" and then create a label with a "•" inside the cell.

The only reason we'd want something like this to live in the view model would be if we were changing the RP indicator year-to-year or something.

the-blue-alliance-ios/View Elements/Match/MatchViewModel.swift
```diff
@@ -44,6 +47,29 @@ struct MatchViewModel {
             }
             return true
         }()
+        
+        // Set RP things, specifically for 2018
+        // Other years can use a similar pattern
+        if match.year == 2018 {
```
ZachOrr: API request: Standardize this so we don't have to update it every year

.gitignore
```diff
@@ -58,3 +58,6 @@ tba-unit-tests/FailureDiffs
 # Ignore some Metadata from Fastlane with sensitive information
 fastlane/metadata/trade_representative_contact_information
 fastlane/metadata/review_information
+
+# rbenv
+.ruby-version
```
ZachOrr: What's this about?

dracco1993: It's used by `rbenv`, a `ruby` virtual environment manager; like `python`'s `virtualenv`.  The `.ruby-version` file specifically tells `rbenv` what version of `ruby` to use for the given folder.
It's essentially a required thing for `ruby` devs.  The normal thing to do is just ignore the file, but if that's an issue I can take it out.

ZachOrr: Gotcha - if we sync about this, we could update the setup docs to make sure everyone is using the same version of Ruby. That seems ideal.

ZachOrr:
We should also support 2016 and 2017

https://github.com/the-blue-alliance/the-blue-alliance-android/pull/866/files#diff-39aebbd26caffc9bc10a5befc1fc27a2R141

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -115,12 +133,26 @@ class MatchSummaryView: UIView {
             blueScoreLabel.font = winnerFont
         }
     }
+    
+    private func addRPToView(stackView: UIStackView, rpCount: Int) {
+        for _ in 0..<rpCount {
+            let rpLabel = label(text: "•", isBold: true)
+            stackView.insertArrangedSubview(rpLabel, at: 0)
```
ZachOrr: I think we can do `addArrangedSubview` here - I'm actually fairly certain the way we're inserting the team labels right now is causing a bug that I've worked around improperly...

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -91,18 +102,25 @@ class MatchSummaryView: UIView {
         playIconImageView.isHidden = viewModel.hasVideos
 
         removeTeams()
+        removeRPs()
 
         for teamKey in viewModel.redAlliance {
-            let teamLabel = label(for: teamKey, baseTeamKey: viewModel.baseTeamKey)
-            redStackView.insertArrangedSubview(teamLabel, at: 0)
+            let redTeamLabel = teamLabel(for: teamKey, baseTeamKey: viewModel.baseTeamKey)
+            redStackView.insertArrangedSubview(redTeamLabel, at: 0)
```
ZachOrr: Can we do `addArrangedSubview` instead of inserting and passing an index?

the-blue-alliance-ios/View Elements/Match/MatchViewModel.swift
```diff
@@ -44,15 +47,53 @@ struct MatchViewModel {
             }
             return true
         }()
-
+
```
ZachOrr: Let's get rid of these please. Sorry - I don't have a linter or anything. I am our linter.

ZachOrr:
If we can add tests (unit tests, at least) and fix those stray spaces, this looks god to go 👍

tba-unit-tests/View Elements/Match/MatchViewModelTests.swift
```diff
@@ -0,0 +1,138 @@
+import XCTest
+@testable import The_Blue_Alliance
+
+class MatchViewModelTestCase: CoreDataTestCase {
+    func test_no_breakdown() {
```
ZachOrr: space above please

tba-unit-tests/View Elements/Match/MatchViewModelTests.swift
```diff
@@ -0,0 +1,138 @@
...
+            "blue": [
+                "teleopDefensesBreached": true,
+                "teleopTowerCaptured": true
+            ]
+        ]
+
+        let subject = MatchViewModel(match: match)
+        XCTAssertEqual(subject.redRPCount, 2)
+        XCTAssertEqual(subject.blueRPCount, 2)
+    }
```
ZachOrr: space after here please

tba-unit-tests/View Elements/Match/MatchViewModelTests.swift
```diff
@@ -0,0 +1,138 @@
...
+}
```
ZachOrr: Can we add a test where we pass the `calculateRP` keys it can't find in the breakdown dictionary?

tba-unit-tests/View Elements/Match/MatchViewModelTests.swift
dracco1993: Added in 9f02a264bb7da94be1164f8501639943760201ef

tba-unit-tests/View Elements/Match/MatchViewModelTests.swift
dracco1993: Added in 9f02a264bb7da94be1164f8501639943760201ef

tba-unit-tests/View Elements/Match/MatchViewModelTests.swift
dracco1993: Added 9f02a264bb7da94be1164f8501639943760201ef

PR #410, opened 2019-01-01
Fix double register call

### LINKED ISSUE #392: Fix double `/register` calls when signing in
When signing in, we don't need https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/master/the-blue-alliance-ios/App%20Delegate/AppDelegate.swift#L347 - this gets called because we get an FCM token. I forget exactly why we added this call, but it should be removed and we should let Firebase manage giving us a token and the PushService will manage registering that token.

ZachOrr:
Closes #392


ZachOrr:
No idea why this build is failing on CI - will kick in the morning

PR #449, opened 2019-02-26
Local Podspec for TBAKit/TBAKitTesting

### LINKED ISSUE #409: Split TBAKit in to monorepo framework
This will be important if we start building extensions. But also - I'm looking to get these tests out of the `tba-unit-test` target

We can probably do this with a slightly more complicated private pod setup https://guides.cocoapods.org/making/private-cocoapods.html

ZachOrr:
Outstanding things left to be done
* Run TBAKit tests on CI
* Revert Podfile.lock changes

Closes #409


ZachOrr:
I *think* this is good to go. It's hard to tell, since the tests can't run on CI.

ZachOrr:
Worth thinking about if we should remove the network classes from `TBAKitTesting` and pull them in to something else? Seems not super important - things can depend on `TBAKitTesting` to use them, but I don't know if that feels good

the-blue-alliance-ios/Frameworks/TBAKit/Testing/MockTBAKit.swift
```diff
@@ -1,12 +1,14 @@
...
 
-class MockTBAKit: TBAKit {
+public class MockTBAKit: TBAKit {
 
-    let session: MockURLSession
+    public let session: MockURLSession
+    private let bundle: Bundle
 
-    init(userDefaults: UserDefaults) {
+    public init(userDefaults: UserDefaults, bundle: Bundle) {
```
ZachOrr: This is problematic... at least, it's failing our tests in the TBA unit tests. I think we might consider having some fake responses we use for these inside of TBA and move on. Feels easiest, and also makes sense. It allows us to control responses between the TBAKit tests and the TBA tests

ZachOrr:
This one is ready to go, but still feels so bad. I need to figure out

* Why we're getting duplicated resources from the testing data (maybe that whole data folder is getting included?)
* How to dynamically add local pod schemes to TBA workspace on CI

ZachOrr:
Finally got this one looking good... needs some docs about how to work on things in isolation, but overall, seems great! MyTBAKit and the Core Data models are next.

PR #475, opened 2019-03-04
add calculated RP field to left column of EventRankings

cyocom:
Adds total RP to the left column of EventRankings for years where it is a calculated field.
Adds RP/m to the left column of EventRankings for years where it is a calculated field.

## Description
adds rpString and rpTotal variables to RankingCellViewModel, computed by finding "Total Ranking Points", or "Qual Score/Match" (rpString) in extraStatsNames, and using the index of that to determine the index of the Total RP in extraStatsValues and assigning that to rpTotal.
prefixes tiebreakerInfoString with ```rpString: rpTotal,```

## Motivation and Context
Data was missing in the app compared to the site, requiring both to have full information.

## How Has This Been Tested?
Tested using the iOS simulator on iPhone X, XR, and 5s

## Screenshots

screen shot 2019-03-04 at 1 34 32 am
screen shot 2019-03-04 at 1 34 41 am
screen shot 2019-03-04 at 1 35 40 am
screen shot 2019-03-04 at 1 35 53 am
screen shot 2019-03-04 at 1 37 56 am

### DIFF
--- a/the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
+++ b/the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
@@ -113,7 +113,7 @@
     <entity name="EventRanking" representedClassName="EventRanking" syncable="YES" codeGenerationType="class">
         <attribute name="dq" optional="YES" attributeType="Integer 16" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="extraStatsNames" optional="YES" attributeType="Transformable" customClassName="[String]" syncable="YES"/>
-        <attribute name="extraStatsValues" optional="YES" attributeType="Transformable" customClassName="[Int]" syncable="YES"/>
+        <attribute name="extraStatsValues" optional="YES" attributeType="Transformable" customClassName="[NSNumber]" syncable="YES"/>
         <attribute name="matchesPlayed" optional="YES" attributeType="Integer 16" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="qualAverage" optional="YES" attributeType="Double" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="rank" attributeType="Integer 16" usesScalarValueType="NO" syncable="YES"/>
--- a/the-blue-alliance-ios/View Elements/Events/RankingCellViewModel.swift
+++ b/the-blue-alliance-ios/View Elements/Events/RankingCellViewModel.swift
@@ -9,8 +9,6 @@ struct RankingCellViewModel {
 
     let detailText: String?
     let wltText: String?
-    
-    let rpText: String?
 
     init(districtRanking: DistrictRanking) {
         rankText = "Rank \(districtRanking.rank!.stringValue)"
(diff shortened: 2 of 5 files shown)


the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
```diff
@@ -113,7 +113,7 @@
     <entity name="EventRanking" representedClassName="EventRanking" syncable="YES" codeGenerationType="class">
         <attribute name="dq" optional="YES" attributeType="Integer 16" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="extraStatsNames" optional="YES" attributeType="Transformable" customClassName="[String]" syncable="YES"/>
-        <attribute name="extraStatsValues" optional="YES" attributeType="Transformable" customClassName="[Int]" syncable="YES"/>
+        <attribute name="extraStatsValues" optional="YES" attributeType="Transformable" customClassName="[NSNumber]" syncable="YES"/>
```
ZachOrr: Are we going to have to write custom migrations for these, or does the `customClassName` stuff work automagically and since the underlying data hasn't changed (I think)

ZachOrr: Looking at tests for this as well - we're using all Int's. We should try to add some test coverage to make sure we support different types.

the-blue-alliance-ios/View Elements/Events/RankingCellViewModel.swift
```diff
@@ -35,12 +35,46 @@ struct RankingCellViewModel {
 
         teamNumber = String(describing: eventRanking.teamKey!.teamNumber)
         teamName = eventRanking.teamKey!.team?.nickname ?? eventRanking.teamKey!.name
+
```
ZachOrr: Not sure if this is outdated but - re: conversation in Slack, shouldn't this code be in `tiebreakerInfoString` ?

cyocom: should be fixed in 531ab95

the-blue-alliance-ios/Core Data/TBA.xcdatamodeld/TBA.xcdatamodel/contents
cyocom: I can look into tests, but I have no idea how to identify / test for issues with core data migrations. If you could point me in the right direction that'd be great 👍

cyocom: actually unsure how to add tests around this as well, mostly because I think this logic should be moved to EventRanking.swift instead of RankingCellViewModel.swift

ZachOrr: Re: Core Data migration - the easiest way to test is to launch the app from master, load some data (go to a rankings view), then checkout your branch and build/run again. If the app crashes on launch, we're gonna have a problem.

I can test this too - I tried to last night, but your branch didn't build for me. Then I got distracted...

ZachOrr:
I'm gonna close this, because I handled this in https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/496 - sorry, I wanted to get this in

Issue #352, opened 2018-12-02
Speed up CI tests


ZachOrr:
This is a weird one but - CI tests are crazy slow. ~10min turnaround time to land a PR is way too long for an app of this size.

We might be able to speed this one up by checking what configuration we're building when running scan. Offhand I'm not sure what else we can do to speed up builds. Most of the time spent seems to be compiling frameworks and stuff.

ZachOrr:
We should double check our caches for deps are working properly (I'm not entirely sure they are)

ZachOrr:
We might be able to pre-boot our simulators as well

ZachOrr:
`skip_build` didn't seem to speed things up

ZachOrr:
We might actually look to move from Cocoapods to something like Carthage where we just install compiled frameworks instead of compiling sources... that feels like it should speed things up. I'm far less familiar with the tooling for Carthage though, especially if we get in to a situation where we want to keep TBAKit/MyTBA/others as sub-projects inside of a monorepo

ZachOrr:
One thing we could do too is add our derived data to our caches (or something) to keep derived data around. Then we could blow our derived data during a deploy to make sure it's fresh. That should speed things up.

ZachOrr:
Here we go - https://github.com/leavez/cocoapods-binary

ZachOrr:
I'm not thinking about this right - the problem with binary pods is we still need to build everything. It'll only speed up subsequent builds. Probably not super helpful... We'd need someone to distribute frameworks and then pull in those to actually get a speedup

PR #478, opened 2019-03-04
Run pod check to speed up CI

ZachOrr:
From http://guides.cocoapods.org/plugins/optimising-ci-times.html

See if we should use `--ignore-dev-pods`


ZachOrr:
Aye this actually looks good - will probably merge once tests pass. I think the trick was to make sure we `pod install` -> `pod install --repo-update` on failure

ZachOrr:
Closes #352 for now

PR #492, opened 2019-03-09
Remove spaces from scheme/app

ZachOrr:
I'm pretty sure the beta builds are failing on the RN build phase because of the spaces in our scheme/app. I have no idea what would cause this regression, but it's worth trying.


ZachOrr:
I (probably) won't be merging this but the `TBA.xcarchive` removal should make it's way to master

PR #502, opened 2019-03-11
Convert EventRanking tiebraker to EventRankingStat/EventRankingStatsInfo

ZachOrr:
Convert the old `extraStatsNames`/`extraStatsValues` and `tiebreakerNames`/`tiebreakerValues` to `extraStatsInfo`/`extraStats` and `sortOrdersInfo`/`sortOrders`, and make them relationships on the model to `EventRankingStat` and `EventRankingStatsInfo` objects instead of string arrays.

The new `EventRankingStatsInfo` model stores precision as well as name, which means our `rankingInfoString` will always show the right precision when display these stats. The `EventRankingStat` is because storing the value on the `EventRankingStatsInfo`, or having some `EventRankingStat` that referred to a `EventRankingStatsInfo` and doing the book keeping for cleanup got WILDLY complicated. We have to keep two ordered arrays in sync, but we get this for "free" with `updateToManyRelationship`, so it's not a big deal.

### DIFF
--- a/the-blue-alliance-ios/Core Data/Event/EventRanking.swift
+++ b/the-blue-alliance-ios/Core Data/Event/EventRanking.swift
@@ -4,23 +4,20 @@ import TBAKit
 
 extension EventRanking: Managed {
 
-    var extraStatsInfoArray: [EventRankingStatInfo]? {
-        return extraStatsInfo?.array as? [EventRankingStatInfo]
+    var extraStatsInfoArray: [EventRankingStatInfo] {
+        return extraStatsInfo?.array as? [EventRankingStatInfo] ?? []
     }
-    var extraStatsArray: [EventRankingStat]? {
-        return extraStats?.array as? [EventRankingStat]
+    var extraStatsArray: [EventRankingStat] {
+        return extraStats?.array as? [EventRankingStat] ?? []
     }
-    var sortOrdersInfoArray: [EventRankingStatInfo]? {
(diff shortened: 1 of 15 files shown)


the-blue-alliance-ios/Core Data/Event/EventRanking.swift
```diff
@@ -94,6 +111,46 @@ extension EventRanking: Managed {
                 qualStatus.ranking = nil
             }
         }
+
+        // Loop twice - once to cleanup our relationships, once to delete.
+        // Note to Zach: Test this with only one loop and confirm it doesn't work
+        extraStatsInfoArray?.forEach({
```
ZachOrr: DRY this, if we can

the-blue-alliance-ios/Core Data/Event/EventRankingStatInfo.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import CoreData
+
+extension EventRankingStatInfo: Managed {
+
+    class func insert(_ model: TBAEventRankingSortOrder, in context: NSManagedObjectContext) -> EventRankingStatInfo {
+        let predicate = NSPredicate(format: "%K == %@ && %K == %ld",
```
ZachOrr: Maybe just query by name here - seems possible we'd update the precision but keep the name upstream

Issue #464, opened 2019-03-02
Make teams in Match VC tappable


ZachOrr:

kevinbai0:
I can tackle this issue.

PR #528, opened 2019-04-29
Finished issue #464

kevinbai0:
Clicking on a team in the `MatchSummaryView` will take them to their `Team@Event` page. Labels were changed to buttons on the MatchSummaryView (when not in the MatchesViewController) and color of the buttons were changed from `Black` to `Primary Blue`.


the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +156,67 @@ class MatchSummaryView: UIView {
...
+        let label = UILabel()
+        label.textAlignment = .center
+        label.attributedText = customAttributedString(text: text, isBold: isBold, isStrikethrough: isStrikethrough)
+        label.translatesAutoresizingMaskIntoConstraints = false
+
+        return label
+    }
+    
+    @objc private func teamPressed(sender: UIButton) {
+        guard let text = sender.titleLabel?.text else { return }
```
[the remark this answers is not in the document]
ZachOrr: Not *exactly* the same (since we're indexing in to lists, as opposed to fetching teams) but here's some examples of what I'm talking about, where we'll set the tag on a view and use it later to get some data

https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/master/the-blue-alliance-ios/View%20Elements/Events/EventAllianceTableViewCell.swift#L91 

https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/master/the-blue-alliance-ios/View%20Elements/Awards/AwardTableViewCell.swift#L65

Gemfile.lock
```diff
@@ -214,4 +214,4 @@ DEPENDENCIES
   fastlane
 
 BUNDLED WITH
-   1.17.1
+   2.0.1
```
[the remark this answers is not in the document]
kevinbai0: Yeah, I didn't realize I had updated the version.

the-blue-alliance-ios/View Controllers/Events/Event/EventViewController.swift
```diff
@@ -121,10 +123,29 @@ extension EventViewController: EventRankingsViewControllerDelegate {
 }
 
 extension EventViewController: MatchesViewControllerDelegate {
-
     func matchSelected(_ match: Match) {
-        let matchViewController = MatchViewController(match: match, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
-        self.navigationController?.pushViewController(matchViewController, animated: true)
+        self.matchViewController = MatchViewController(match: match, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
+        self.matchViewController?.matchSummaryDelegate = self
```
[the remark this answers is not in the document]
kevinbai0: I couldn't seem to find the TeamKey directly in MatchViewController but I'll try and find it.

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -1,13 +1,22 @@
 import Foundation
 import UIKit
 
-class MatchSummaryView: UIView {
+protocol MatchSummaryViewDelegate:class {
+    func teamPressed(teamNumber: Int)
```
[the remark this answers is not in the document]
kevinbai0: I'll change this method.

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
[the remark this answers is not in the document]
kevinbai0: Tags would be better but we could also create an extensible tappable UIButton like you said that contains TeamKeys themselves which would also solve this issue.

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +156,67 @@ class MatchSummaryView: UIView {
...
+    
+    private func button(text: String, isBold: Bool, isStrikethrough: Bool = false, isTeam: Bool = true) -> UIButton {
+        let newButton = UIButton(type: .system)
+        newButton.setTitle(text, for: [])
+        
+        // set color of the label to black like default
+        let buttonColor = UIColor.primaryBlue
+        newButton.setTitleColor(buttonColor, for: .normal)
+        // change color to gray when user hits the button
+        newButton.setTitleColor(buttonColor.withAlphaComponent(0.8), for: .highlighted)
```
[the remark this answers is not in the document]
kevinbai0: The default color is a light blue, which didn't fit any color that's currently used in the app. It shouldn't be black since there needs to be an indicator that it is tappable. I chose the primary blue color that's used throughout the app to avoid adding in a completely new color but we can definitely change that accordingly.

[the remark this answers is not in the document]
ZachOrr: The primary blue for the base color seems fine - but is there a reason we need to override the tint color stuff? Why not let the system handle that by default?

the-blue-alliance-ios/View Controllers/Events/Event/EventViewController.swift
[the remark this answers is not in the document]
ZachOrr: You should be able to get team key objects from alliances on matches - `match.redAlliance.teams` and `match.blueAlliance.teams` will get you array of team keys for both alliances.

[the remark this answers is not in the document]
kevinbai0: Also, Team@Event VC needs a `StatusService` and `UrlOpener`. I'm not too sure what they do but they seem to be passed down to the EventViewController. `MatchViewController` doesn't have one passed down to it by default. Should local properties of `StatusService` and `UrlOpener` be added to MatchViewController or should they be added to its parent (`ContainerViewController`)?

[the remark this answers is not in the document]
ZachOrr: We should add those two properties in the initializer of the `MatchViewController`. We pass these objects around as part of our dependency injection.

the-blue-alliance-ios/View Controllers/Events/Event/EventViewController.swift
```diff
@@ -121,10 +123,8 @@ extension EventViewController: EventRankingsViewControllerDelegate {
 }
 
 extension EventViewController: MatchesViewControllerDelegate {
-
     func matchSelected(_ match: Match) {
-        let matchViewController = MatchViewController(match: match, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
-        self.navigationController?.pushViewController(matchViewController, animated: true)
+        self.matchViewController = MatchViewController(match: match, statusService: statusService, urlOpener: urlOpener, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
```
ZachOrr: We shouldn't have to hold on to this reference anymore, right? We should be able to remove this and the private property above

Gemfile.lock
```diff
@@ -58,10 +58,10 @@ GEM
     digest-crc (0.4.1)
     domain_name (0.5.20180417)
       unf (>= 0.0.5, < 1.0.0)
-    dotenv (2.7.1)
+    dotenv (2.7.2)
```
ZachOrr: Are we bumping these dependencies for a reason? Or is this just accidental?

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +43,38 @@ class MatchViewController: MyTBAContainerViewController {
             tbaKit: tbaKit,
             userDefaults: userDefaults
         )
+        self.infoViewController.matchSummaryDelegate = self
```
ZachOrr: nit: can we do a newline after the init, and then set the delegate?  (we usually leave some space https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/master/the-blue-alliance-ios/View%20Controllers/Events/EventsContainerViewController.swift#L31)

Also, we can drop the `self.` off of this and just set the delegate by doing `infoViewController.matchSummaryDelegate`

the-blue-alliance-ios/View Controllers/Match/MatchInfoViewController.swift
```diff
@@ -61,6 +61,15 @@ class MatchInfoViewController: TBAViewController, Observable {
         videoStackView.translatesAutoresizingMaskIntoConstraints = false
         return videoStackView
     }()
+    
+    public var matchSummaryDelegate: MatchSummaryViewDelegate? {
+        get {
+            return self.matchSummaryView.delegate
```
ZachOrr: I think we can drop the `self` from these as well?

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +43,38 @@ class MatchViewController: MyTBAContainerViewController {
...
     }
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-
+
```
ZachOrr: nit: white space

the-blue-alliance-ios/View Controllers/Match/MatchesViewController.swift
```diff
@@ -65,7 +65,6 @@ class MatchesViewController: TBATableViewController {
     override func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
         return 30.0
     }
-
```
ZachOrr: nit: can we back out this change?

the-blue-alliance-ios/View Controllers/Teams/Team/Team@Event/TeamAtEventViewController.swift
```diff
@@ -87,8 +88,8 @@ extension TeamAtEventViewController: MatchesViewControllerDelegate, TeamSummaryV
     }
 
     func matchSelected(_ match: Match) {
-        let matchViewController = MatchViewController(match: match, teamKey: teamKey, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
-        self.navigationController?.pushViewController(matchViewController, animated: true)
+        self.matchViewController = MatchViewController(match: match, statusService: statusService, urlOpener: urlOpener, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
```
ZachOrr: Same comment as above - we should be able to remove this code now

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +157,67 @@ class MatchSummaryView: UIView {
     private func teamLabel(for teamKey: String, baseTeamKey: String?, dq: Bool) -> UILabel {
         let text: String = "\(Team.trimFRCPrefix(teamKey))"
         let isBold: Bool = (teamKey == baseTeamKey)
-
+
```
ZachOrr: nit: white space

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +157,67 @@ class MatchSummaryView: UIView {
...
         let attributedStringRange = NSMakeRange(0, attributeString.length)
-
+        
         var font: UIFont = .systemFont(ofSize: 14)
         if isBold {
             font = .boldSystemFont(ofSize: 14)
         }
         attributeString.addAttribute(.font, value: font, range: attributedStringRange)
-
+
```
ZachOrr: nit: white space

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +157,67 @@ class MatchSummaryView: UIView {
...
+    @objc private func teamPressed(sender: UIButton) {
+        if sender.tag == 0 { return }
+        self.delegate?.teamPressed(teamNumber: sender.tag)
+    }
+    
+    private func customAttributedString(text: String, isBold: Bool, isStrikethrough: Bool = false) -> NSMutableAttributedString {
         let attributeString =  NSMutableAttributedString(string: text)
         let attributedStringRange = NSMakeRange(0, attributeString.length)
-
+
```
ZachOrr: nit: white space

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +157,67 @@ class MatchSummaryView: UIView {
...
+        label.textAlignment = .center
+        label.attributedText = customAttributedString(text: text, isBold: isBold, isStrikethrough: isStrikethrough)
+        label.translatesAutoresizingMaskIntoConstraints = false
+
+        return label
+    }
+    
+    @objc private func teamPressed(sender: UIButton) {
+        if sender.tag == 0 { return }
+        self.delegate?.teamPressed(teamNumber: sender.tag)
```
ZachOrr: I think we can drop the `self` from this

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +157,67 @@ class MatchSummaryView: UIView {
...
+    }
+    
+    private func button(text: String, isBold: Bool, isStrikethrough: Bool = false, isTeam: Bool = true) -> UIButton {
+        let newButton = UIButton(type: .system)
+        newButton.setTitle(text, for: [])
+        
+        // set color of the label to black like default
+        newButton.setTitleColor(.primaryBlue, for: .normal)
+        
+        if isTeam {
```
ZachOrr: Can we combine this with the next line? We should be able to say

`if isTeam, if let teamNumber = Int(text) {`

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -144,30 +157,67 @@ class MatchSummaryView: UIView {
...
     }
+    
+    private func teamButton(for teamKey: String, baseTeamKey: String?, dq: Bool) -> UIButton {
+        let text: String = "\(Team.trimFRCPrefix(teamKey))"
+        let isBold: Bool = (teamKey == baseTeamKey)
+
+        return button(text: text, isBold: isBold, isStrikethrough: dq)
+    }
+    
+    private func button(text: String, isBold: Bool, isStrikethrough: Bool = false, isTeam: Bool = true) -> UIButton {
```
ZachOrr: I might be missing it - is there someplace we use this method that `isTeam` is `false`?

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +43,38 @@ class MatchViewController: MyTBAContainerViewController {
...
         Analytics.logEvent("match", parameters: ["match": match.key!])
     }
+    
+}
 
+extension MatchViewController: MatchSummaryViewDelegate {
+    
+    func teamPressed(teamNumber: Int) {
+        guard let event = match.event else { return }
+        guard let redTeamKeys = match.redAlliance?.teams?.array as? [TeamKey] else { return }
```
ZachOrr: @ZachOrr Add a computed property on `Match` that aggregates all teams in a match. That way we can simplify this and do `match.teams`

ZachOrr: I also think we can tighten up this logic a little using Swift but I'm not entirely sure how yet... will look this evening when I'm at a compiler

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +43,38 @@ class MatchViewController: MyTBAContainerViewController {
...
+        guard let blueTeamKeys = match.blueAlliance?.teams?.array as? [TeamKey] else { return }
+        
+        // get team key that matches the target teamNumber
+        guard let teamKey = (redTeamKeys + blueTeamKeys).filter({
+            $0.teamNumber == "\(teamNumber)"
+        }).first else { return }
+        
+        let teamAtEventVC = TeamAtEventViewController(teamKey: teamKey, event: event, myTBA: self.myTBA, showDetailEvent: true, showDetailTeam: true, statusService: self.statusService, urlOpener: self.urlOpener, persistentContainer: self.persistentContainer, tbaKit: self.tbaKit, userDefaults: self.userDefaults)
+        self.navigationController?.pushViewController(teamAtEventVC, animated: true)
+
```
ZachOrr: nit: remove newline

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +43,38 @@ class MatchViewController: MyTBAContainerViewController {
...
+        guard let redTeamKeys = match.redAlliance?.teams?.array as? [TeamKey] else { return }
+        guard let blueTeamKeys = match.blueAlliance?.teams?.array as? [TeamKey] else { return }
+        
+        // get team key that matches the target teamNumber
+        guard let teamKey = (redTeamKeys + blueTeamKeys).filter({
+            $0.teamNumber == "\(teamNumber)"
+        }).first else { return }
+        
+        let teamAtEventVC = TeamAtEventViewController(teamKey: teamKey, event: event, myTBA: self.myTBA, showDetailEvent: true, showDetailTeam: true, statusService: self.statusService, urlOpener: self.urlOpener, persistentContainer: self.persistentContainer, tbaKit: self.tbaKit, userDefaults: self.userDefaults)
+        self.navigationController?.pushViewController(teamAtEventVC, animated: true)
```
ZachOrr: I think we can drop the `self` from here - I know I have a lot of places where I call `self.navigationController`, I write a lot of Obj-C still 😛

Gemfile.lock
kevinbai0: Completely accidental, I'm not too familiar with this build environment.

the-blue-alliance-ios/Core Data/Match/Match.swift
```diff
@@ -169,6 +169,15 @@ extension Match {
     var blueAllianceTeamNumbers: [String] {
         return blueAllianceTeamKeys.map({ Team.trimFRCPrefix($0) })
     }
+    
+    /**
+     Returns the team keys from both alliances
+    */
+    var teamKeys: [TeamKey]? {
+        guard let blueKeys = blueAlliance?.teams?.array as? [TeamKey] else { return nil }
```
ZachOrr: I think we can do better here - use the `alliances` property, which looks like a set of `MatchAlliance` objects. And then we should be able to flatten the two sets in to one.

Also, these methods usually return empty objects instead of nulls - see `dqTeamKeys` for example.

the-blue-alliance-ios/Core Data/Match/Match.swift
```diff
@@ -169,6 +169,15 @@ extension Match {
     var blueAllianceTeamNumbers: [String] {
         return blueAllianceTeamKeys.map({ Team.trimFRCPrefix($0) })
     }
+    
+    /**
+     Returns the team keys from both alliances
```
ZachOrr: nit: `Returns the team keys from both alliances` -> `Returns the team keys for all alliances` once we implemented feedback of using `alliances`

the-blue-alliance-ios/Core Data/Match/Match.swift
ZachOrr: Also - can we make sure this gets tests? All the Core Data stuff is pretty well tested - I wanna make sure we keep test coverage high on the models.

the-blue-alliance-ios/View Controllers/Events/Event/EventViewController.swift
```diff
@@ -121,10 +121,8 @@ extension EventViewController: EventRankingsViewControllerDelegate {
 }
 
 extension EventViewController: MatchesViewControllerDelegate {
-
```
ZachOrr: nit: please don't remove whitespace

the-blue-alliance-ios/View Controllers/Events/Event/EventViewController.swift
```diff
@@ -121,10 +121,8 @@ extension EventViewController: EventRankingsViewControllerDelegate {
...
     func matchSelected(_ match: Match) {
-        let matchViewController = MatchViewController(match: match, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
+        let matchViewController = MatchViewController(match: match, statusService: statusService, urlOpener: urlOpener, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
         self.navigationController?.pushViewController(matchViewController, animated: true)
     }
-
```
ZachOrr: nit: please don't remove whitespace

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -10,18 +10,21 @@ class MatchViewController: MyTBAContainerViewController {
...
     // MARK: Init
 
-    init(match: Match, teamKey: TeamKey? = nil, myTBA: MyTBA, persistentContainer: NSPersistentContainer, tbaKit: TBAKit, userDefaults: UserDefaults) {
+    init(match: Match, teamKey: TeamKey? = nil, statusService: StatusService, urlOpener: URLOpener, myTBA: MyTBA, persistentContainer: NSPersistentContainer, tbaKit: TBAKit, userDefaults: UserDefaults) {
         self.match = match
-
+        self.statusService = statusService
+        self.urlOpener = urlOpener
         infoViewController = MatchInfoViewController(match: match, teamKey: teamKey, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
-
```
ZachOrr: nit: please don't remove whitespace

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +42,35 @@ class MatchViewController: MyTBAContainerViewController {
...
         )
+        
+        infoViewController.matchSummaryDelegate = self
     }
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-
+
```
ZachOrr: nit: whitespace

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -39,18 +42,35 @@ class MatchViewController: MyTBAContainerViewController {
...
         Analytics.logEvent("match", parameters: ["match": match.key!])
     }
+    
+}
 
+extension MatchViewController: MatchSummaryViewDelegate {
+    
+    func teamPressed(teamNumber: Int) {
+        guard let event = match.event,
+            let teamKeys = match.teamKeys else { return }
```
ZachOrr: nit: This should be able to be one line, right? Length doesn't seem like an issue here

the-blue-alliance-ios/View Controllers/Match/MatchesViewController.swift
```diff
@@ -65,7 +65,7 @@ class MatchesViewController: TBATableViewController {
...
+
```
ZachOrr: nit: whitepsace

the-blue-alliance-ios/Core Data/Match/Match.swift
kevinbai0: For the test, how should I go about checking the array of Team Keys? Currently, I just checked to see if the length of the teamKeys is equal to 6 or should I go through each index and see if the whole object of Team Key checks out?

ZachOrr: I would check the keys for sure - here's an example of where we do that in other tests https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/2c4d804e196292309774e9e4cbf3d27ccaa45391/tba-unit-tests/Core%20Data/Match/MatchTests.swift#L477

the-blue-alliance-ios/Core Data/Match/Match.swift
```diff
@@ -169,6 +169,13 @@ extension Match {
     var blueAllianceTeamNumbers: [String] {
         return blueAllianceTeamKeys.map({ Team.trimFRCPrefix($0) })
     }
+    
+    /**
+     Returns the team keys for all alliances
+    */
+    var teamKeys: [TeamKey] {
+        return (alliances?.allObjects as? [MatchAlliance])?.reduce([], { $0 + ($1.teams?.array as? [TeamKey] ?? []) }) ?? []
```
ZachOrr: Hell yeah 🙌

the-blue-alliance-ios.xcodeproj/project.pbxproj
```diff
@@ -1739,6 +1739,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
```
ZachOrr: Oh noooo

the-blue-alliance-ios/View Controllers/Match/MatchViewController.swift
```diff
@@ -52,5 +58,19 @@ class MatchViewController: MyTBAContainerViewController {
...
 
+extension MatchViewController: MatchSummaryViewDelegate {
+    
+    func teamPressed(teamNumber: Int) {
+        guard let event = match.event else { return }
+        
+        // get team key that matches the target teamNumber
+        guard let teamKey = match.teamKeys.first(where: { $0.teamNumber == "\(teamNumber)"}) else { return }
+        
+        let teamAtEventVC = TeamAtEventViewController(teamKey: teamKey, event: event, myTBA: myTBA, showDetailEvent: true, showDetailTeam: true, statusService: statusService, urlOpener: urlOpener, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
```
ZachOrr: We need `showDetailEvent` and `showDetailTeam` shouldn't both be true here - we need to figure out which one we want. The right way to do this would be to figure out how we entered the view and show the opposite, but I think we'll just leave it to always push to the Event to keep things simpler.

the-blue-alliance-ios/View Controllers/Teams/Team/Team@Event/TeamAtEventViewController.swift
```diff
@@ -87,7 +86,7 @@ extension TeamAtEventViewController: MatchesViewControllerDelegate, TeamSummaryV
     }
 
     func matchSelected(_ match: Match) {
-        let matchViewController = MatchViewController(match: match, teamKey: teamKey, myTBA: myTBA, persistentContainer: persistentContainer, tbaKit: tbaKit, userDefaults: userDefaults)
```
ZachOrr: Oops missed this - we need to pass the teamKey still. This broke the bolding logic when testing.

PR #530, opened 2019-05-02
Stop hiding nav bar when pushing to a new view controller

ZachOrr:
This was a suggestion - I figure we can try it out and see how we like it


BK1031:
Yeah, I prefer this a lot more than before. It also follows Apple's Human Interface Guidelines.

ZachOrr:
I’m not sure I’ve ever seen “don’t hide the tab bar” in the HIG - where is this guidance?

BK1031:
super late but I just found it lol
https://developer.apple.com/design/human-interface-guidelines/ios/bars/tab-bars/

Screen Shot 2019-09-08 at 8 53 09 PM

PR #536, opened 2019-05-05
[WIP] Team header with avatar

ZachOrr:
Simulator Screen Shot - iPhone Xs - 2019-05-10 at 21 14 59


ZachOrr:
We need to fix the button padding https://noahgilmore.com/blog/uibutton-padding
Also I think our header needs to slide better kinda like https://michiganlabs.com/ios/development/2016/05/31/ios-animating-uitableview-header/

Issue #336, opened 2018-11-30
Add `notModified` boolean to TBAKit callback


ZachOrr:
We should look to add a `notModified` boolean to the TBAKit callback blocks when we get a 304 back from the API - this should help us figure out what to do with a request (should we delete the existing objects or leave them alone)

ZachOrr:
Fixed by https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/538

PR #538, opened 2019-05-05
Convert TBAKit to return Result types

ZachOrr:
Result types, baby.


the-blue-alliance-ios/Frameworks/TBAKit/Sources/TBAKit.swift
```diff
@@ -115,61 +115,61 @@ open class TBAKit: NSObject {
...
-                models = []
-                for result in json {
-                    if let model = T(json: result) {
-                        models!.append(model)
-                    }
-                }
+            if let error = error {
+                completion(.failure(error))
+            } else if let statusCode = response?.statusCode, statusCode == 304 {
+                completion(.success([]))
```
ZachOrr: This is going to be problematic - an empty (null) response is different then a successful empty response and a cached empty response.

I thought https://github.com/the-blue-alliance/the-blue-alliance-ios/issues/336 got closed out, but I guess not. We should just throw this boolean on the responses. Also - we'll need to send etag information for Cloudflare, but that's already half done in dev.

Anywhere we're calling `setLastModified` directly we probably shouldn't. But it's kinda not worth passing the whole response object back up the chain to have TBA send it back down to TBAKit... will need to think about this one.

ZachOrr: Also keep the Teams view in mind on this one. We’ll bail early on a refresh if we’re not careful.

the-blue-alliance-ios/View Controllers/MyTBA/MyTBATableViewController.swift
```diff
@@ -265,11 +265,10 @@ class MyTBATableViewController<T: MyTBAEntity & MyTBAManaged, J: MyTBAModel>: TB
     @discardableResult
     func fetchEvent(_ key: String) -> URLSessionDataTask {
         var request: URLSessionDataTask?
-        request = tbaKit.fetchEvent(key: key) { (event, error) in
+        request = tbaKit.fetchEvent(key: key) { (result, notModified) in
             let context = self.persistentContainer.newBackgroundContext()
-
             context.performChangesAndWait({
-                if let event = event {
+                if !notModified, let event = try? result.get() {
```
ZachOrr: I don't think we need to look at `notModified` here

the-blue-alliance-ios/View Controllers/MyTBA/MyTBATableViewController.swift
```diff
@@ -289,10 +288,10 @@ class MyTBATableViewController<T: MyTBAEntity & MyTBAManaged, J: MyTBAModel>: TB
     @discardableResult
     func fetchTeam(_ key: String) -> URLSessionDataTask {
         var request: URLSessionDataTask?
-        request = tbaKit.fetchTeam(key: key) { (team, error) in
+        request = tbaKit.fetchTeam(key: key) { (result, notModified) in
             let context = self.persistentContainer.newBackgroundContext()
             context.performChangesAndWait({
-                if let team = team {
+                if !notModified, let team = try? result.get() {
```
ZachOrr: I don't think we need to look at `notModified` here

the-blue-alliance-ios/View Controllers/MyTBA/MyTBATableViewController.swift
```diff
@@ -312,10 +311,10 @@ class MyTBATableViewController<T: MyTBAEntity & MyTBAManaged, J: MyTBAModel>: TB
     @discardableResult
     func fetchMatch(_ key: String) -> URLSessionDataTask {
         var request: URLSessionDataTask?
-        request = tbaKit.fetchMatch(key: key) { (match, error) in
+        request = tbaKit.fetchMatch(key: key) { (result, notModified) in
             let context = self.persistentContainer.newBackgroundContext()
             context.performChangesAndWait({
-                if let match = match {
+                if !notModified, let match = try? result.get() {
```
ZachOrr: I don't think we need to look at `notModified` here

the-blue-alliance-ios/View Controllers/Settings/SettingsViewController.swift
```diff
@@ -373,6 +373,7 @@ class SettingsViewController: TBATableViewController {
     internal func deleteNetworkCache() {
         userDefaults.clearSuccessfulRefreshes()
         tbaKit.clearLastModified()
+        tbaKit.clearEtag()
```
ZachOrr: Condense these two calls in to one method in TBAKit

the-blue-alliance-ios/View Controllers/Teams/Team/Team@Event/TeamSummaryViewController.swift
```diff
@@ -342,10 +348,10 @@ extension TeamSummaryViewController: Refreshable {
         }
 
         var request: URLSessionDataTask?
-        request = tbaKit.fetchMatch(key: key) { (match, error) in
+        request = tbaKit.fetchMatch(key: key) { (result, notModified) in
             let context = self.persistentContainer.newBackgroundContext()
             context.performChangesAndWait({
-                if let match = match {
+                if !notModified, let match = try? result.get() {
```
ZachOrr: I don't think we need to look at `notModified` here

PR #569, opened 2019-05-09
Fix Core Data threading crashes

### LINKED ISSUE #278: Take pass at Core Data concurrency
We're not respecting Core Data's concurrency all the time - take a pass at it

ZachOrr:
Fix the Core Data threading issues that are causing lots of crashes

Fixes #278

### DIFF
--- a/the-blue-alliance-ios/Core Data/Event/Event.swift
+++ b/the-blue-alliance-ios/Core Data/Event/Event.swift
@@ -381,7 +381,7 @@ extension Event: Locatable, Surfable, Managed {
     }
 
     public func dateString() -> String? {
-        guard let startDate = getValue(\Event.startDate), let endDate = getValue(\Event.endDate) else {
+        guard let startDate = startDate, let endDate = endDate else {
             return nil
         }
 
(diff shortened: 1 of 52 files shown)


the-blue-alliance-ios/Core Data/Event/Event.swift
```diff
@@ -514,11 +520,12 @@ extension Event: Locatable, Surfable, Managed {
     // We need a full object as opposed to a number because of CMP, off-season, etc.
     // TODO: Convert this to a data model that uses a Core Data model for init but isn't a Core Data model
     static func weekEvents(for year: Int, in managedObjectContext: NSManagedObjectContext) -> [Event] {
+        // TODO: This has a problem being called in the background thread - but I'm not entirely sure wy. We need to solve this.
```
ZachOrr: TODO

ZachOrr:
Tests are failing from this LOC

https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/569/files#diff-1efb06396d6c0f01008010c801e12fe9R68

`
test_image, failed: caught "NSInvalidArgumentException", "Unacceptable type of value for attribute: property = "mediaData"; desired type = NSData; given type = NSNull; value = <null>."
`

We should probably kill those setter methods altogether.

Issue #439, opened 2019-02-20
Add Week # to Event cells


ZachOrr:
In some cases (such as viewing Team Events), it isn’t obvious what Week an Event is. We should add something to the Event cell to show what week an Event is, and show it when that data doesn’t exist anywhere else

PR #570, opened 2019-05-10
[WIP] Add week numbers in EventCellView

akmin04:
Added the Week # to EventCellView, making it easier to see what weeks a team competed in.

The label will only show the week for Regional Events and District Events in which the event's `weekString` is "Week 0.5..7". Preseason, championships, and offseason events are usually clear enough.

## Description

- Added `eventWeek` to `EventCellViewModel.swift`

- Added `weekLabel` IBOutlet to `EventTableViewCell.swift` 

- Moved `Name Label` and added `Week Label` to a new `UIStackView` in `EventTableViewCell.xib`

## Motivation and Context
Solves #439 

## How Has This Been Tested?
Tested on iPhone 8+ and iPhone Xs simulator.

## Screenshots

Before:

After:


the-blue-alliance-ios/View Elements/Events/EventCellViewModel.swift
```diff
@@ -3,13 +3,25 @@ import Foundation
...
     init(event: Event) {
         eventShortname = event.safeShortName
         eventLocation = event.locationString
         eventDate = event.dateString()
+        
+        /**
+         Only show the week number if the event if it is in Week 1..7
+         Other events (Offseason, Championships, Preseason) already have their descriptions in the table view.
+         */
+        if event.weekString.contains("Week") {
```
ZachOrr: I don't love this - we could split out the week string calculation section of `weekString` (maybe we call that method `weekString`, and call the existing `weekString` method something else) and call that directly in this case.

the-blue-alliance-ios/View Elements/Events/EventCellViewModel.swift
```diff
@@ -3,13 +3,25 @@ import Foundation
...
+            eventWeek = event.weekString
+        }
+        else {
```
ZachOrr: nit: can we combine this with the line above?

```
if {
...
} else {
...
}
```

the-blue-alliance-ios/View Elements/Events/EventCellViewModel.swift
andykmin: Will rename the current `weekString` to `title` and make a new `weekString`

the-blue-alliance-ios/Core Data/Event/Event.swift
```diff
@@ -448,6 +448,17 @@ extension Event: Locatable, Surfable, Managed {
         }
     }
 
+    /**
+     Same as title but return nil if title isn't in Week 0.5..7
+     */
+    public var weekString: String? {
+        if title.contains("Week") {
```
ZachOrr: We shouldn't need to do any substring matching here - we have the logic already to calculate a week string based on Event properties. We can pull it out of the previous `weekString` method and use it here

the-blue-alliance-ios/Core Data/Event/Event.swift
```diff
@@ -402,7 +402,7 @@ extension Event: Locatable, Surfable, Managed {
         }
     }
 
-    public var weekString: String {
+    public var title: String {
```
ZachOrr: `title` on an Event seems like it would be the name of an event. We might consider naming this better. Maybe something like, `eventGroupString` or `eventGroupingString`?

andykmin:
Anyone know an easy way to add left/right padding on a `UILabel`? Right now the week number labels are being cut off when I add a corner radius.

Right now, I'm adding 2 spaces before and after to create "padding" which is a terrible solution. (Yet somehow works on all accessibility font settings on iPhone Xs)

ZachOrr:
You couldn’t be using artificial padding in the label, you should be using a stack view + auto layout. 

Where the event name label is now you can remove the label, add a horizontal stack view, add a label to the left side and have the other element (another label, yeah?) on the right side. Make sure to set the content hugging properly on the labels so they'll size properly.

andykmin:
Hmm, maybe I'm misinterpreting but that's what I have right now. I set the content hugging so that the week number is lower priority, which makes it look like this:

Screen Shot 2019-05-23 at 9 58 11 PM

However, once I add the corner radius, the edges of the text cut off like this:

Screen Shot 2019-05-23 at 10 00 33 PM

Maybe wrap the week label in a view?

andykmin:
Fixed by subclassing `UILabel` to make `EventWeekLabel`, which just overrides `intrinsicContentSize` to add the labels height to the width (left/right padding). Also adds the corner radius on initialization

ZachOrr:
Wait - you shouldn't have to subclass anything. You can use a stack view with a spacer view that will absorb extra space. This is a rough example of what I'm talking about. The `EventTableViewCell` has a `nameLabel` that now sits in a horizontal stack view with a new label (think of this as your new week label element) and a spacer view that is configured to grow to expand between the space between the title label and the week label.

Screen Shot 2019-05-25 at 9 21 24 AM

Simulator Screen Shot - iPhone Xs - 2019-05-25 at 09 24 56

You can play with the auto layout constraints to get the sizing you want, and play with the stack view spacing to make sure there's enough distance between elements.

andykmin:
How should the spacer cell be used to give the week label padding? I've tried manually setting the width anchor of the week label in `EventTableViewCell` to its width + height to give it "padding"

`weekLabel.widthAnchor.constraint(equalToConstant: weekLabel.frame.width + weekLabel.frame.height).isActive = true`

However, this results in layout constraint errors.

andykmin:
iPhone 5s:


iPhone Xs:


iPad Pro (12.9-inch):

PR #572, opened 2019-05-11
Make Team button in Match view bigger

### LINKED ISSUE #544: Make Team button in Match view bigger
The tap targets here are a little small - they're 30px. We should look to make them 44px. This probably means making the font size bigger. We want to make sure we keep things smaller in the Match list, since we want to display as many as possible above the fold. But we have plenty of room to make things bigger on the Match view.
Simulator Screen Shot - iPhone Xs - 2019-05-05 at 12 00 49

kevinbai0:
Closes #544

### DIFF
--- a/the-blue-alliance-ios/View Controllers/Match/MatchInfoViewController.swift
+++ b/the-blue-alliance-ios/View Controllers/Match/MatchInfoViewController.swift
@@ -117,7 +117,7 @@ class MatchInfoViewController: TBAViewController, Observable {
 
     func styleInterface() {
         scrollView.addSubview(matchStackView)
-        matchStackView.autoSetDimension(.height, toSize: 90)
+        matchStackView.autoSetDimension(.height, toSize: matchSummaryView.calculatedHeight)
         matchStackView.autoPinEdge(.top, to: .top, of: scrollView, withOffset: 8)
         matchStackView.autoPinEdge(toSuperviewSafeArea: .leading, withInset: 16)
         matchStackView.autoPinEdge(toSuperviewSafeArea: .trailing, withInset: 16)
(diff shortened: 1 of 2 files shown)


the-blue-alliance-ios/View Controllers/Match/MatchInfoViewController.swift
```diff
@@ -117,7 +117,7 @@ class MatchInfoViewController: TBAViewController, Observable {
 
     func styleInterface() {
         scrollView.addSubview(matchStackView)
-        matchStackView.autoSetDimension(.height, toSize: 90)
+        matchStackView.autoSetDimension(.height, toSize: matchSummaryView.calculatedHeight)
```
ZachOrr: This is tricky... I think the right way to go about doing this is actually inside of the `MatchSummaryView`, and use constraints to set a minimum size of 44px height on the buttons we generate, and then leave the labels as having whatever height they already have. That way we'll be able to show big buttons where we want buttons, and small text where we want text, and we don't have to mess with the stack view much.

This constraint can probably go away though I hope, if the system has enough information to size itself based on the button heights. Also we'll probably want to change that ` infoStackView.autoMatch(.height, to: .height, of: matchStackView, withMultiplier: (1.0/3.0))` line below. We should just go ahead and set it to be 30px.

the-blue-alliance-ios/View Elements/Match/MatchSummaryView.swift
```diff
@@ -14,12 +14,23 @@ class MatchSummaryView: UIView {
...
+        return (currentFontSize - 14) * multiplier + 90
+    }
 
-    private let winnerFont = UIFont.systemFont(ofSize: 14, weight: UIFont.Weight.bold)
-    private let notWinnerFont = UIFont.systemFont(ofSize: 14, weight: UIFont.Weight.medium)
+    private var currentFontSize: CGFloat {
+        return teamsTappable ? 18 : 14 // change this value to change button size
+    }
+    private var winnerFont: UIFont {
+        return UIFont.systemFont(ofSize: currentFontSize, weight: UIFont.Weight.bold)
```
ZachOrr: Can we see if we can get some dynamic type fonts in here? I can help pick some, if we need.

PR #596, opened 2019-05-23
Assume updateRefresh called on main thread

ZachOrr:
`updateRefresh` should only be called in Refreshable, and it executes on `OperationQueue.main`, so it should only be called on the main thread.


ZachOrr:
Going to close this one for now - it blows, but there's handful of issues I need to fix before I can land this

PR #625, opened 2019-09-25
Bump RN to 0.61.1

ZachOrr:
I'm waiting on some feedback from the RN folks about the ballooning of dependencies in order to prevent a crash during runtime - hopefully they'll be able to help. We shouldn't need to pull in `React-CoreModules` and all of it's dependencies, but we have to in order not to crash. We might also look to patch this version of RN and keep moving


ZachOrr:
Pulling in the template for a base iOS project from https://github.com/facebook/react-native/blob/master/template/ios/Podfile

PR #627, opened 2019-09-28
Fix bad RN merge

ZachOrr:
My RN merge got a little bork'd - this should fix it up and put it back in step with the rn repo


ZachOrr:
I think this is failing due to the Cocoapods update... going to roll back

Issue #664, opened 2019-10-14
Fix crash in AppDelegate.setupGoogleAuthentication


ZachOrr:
Related to changes in 1463aa7700cb51cd429dcd9e394d9d14057dbf4d - we might need to symbolicate some other crash logs to figure out what's going on, or try to replicate in dev.

```
Crashed: com.apple.main-thread
0  The Blue Alliance              0x10072a328 AppDelegate.setupGoogleAuthentication() + 310 (AppDelegate.swift:310)
1  The Blue Alliance              0x10072df44 specialized AppDelegate.application(_:didFinishLaunchingWithOptions:) + 182 (AppDelegate.swift:182)
2  The Blue Alliance              0x1007299d0 @objc AppDelegate.application(_:didFinishLaunchingWithOptions:) (<compiler-generated>)
3  UIKitCore                      0x1c863cebc -[UIApplication _handleDelegateCallbacksWithOptions:isSuspended:restoreState:] + 340
4  UIKitCore                      0x1c863ec8c -[UIApplication _callInitializationDelegatesWithActions:forCanvas:payload:fromOriginatingProcess:] + 5076
5  UIKitCore                      0x1c86444c0 -[UIApplication _runWithMainScene:transitionContext:completion:] + 1304
6  UIKitCore                      0x1c7e20810 -[_UISceneLifecycleMultiplexer completeApplicationLaunchWithFBSScene:transitionContext:] + 148
7  UIKitCore                      0x1c82b13ec _UIScenePerformActionsWithLifecycleActionMask + 100
8  UIKitCore                      0x1c7e212f8 __101-[_UISceneLifecycleMultiplexer _evalTransitionToSettings:fromSettings:forceExit:withTransitionStore:]_block_invoke + 196
9  UIKitCore                      0x1c7e20d50 -[_UISceneLifecycleMultiplexer _performBlock:withApplicationOfDeactivationReasons:fromReasons:] + 296
10 UIKitCore                      0x1c7e21124 -[_UISceneLifecycleMultiplexer _evalTransitionToSettings:fromSettings:forceExit:withTransitionStore:] + 736
11 UIKitCore                      0x1c7e209e0 -[_UISceneLifecycleMultiplexer uiScene:transitionedFromState:withTransitionContext:] + 336
[...]

PR #671, opened 2019-10-15
Revert GoogleSignIn version

ZachOrr:
An attempt to fix users that are crashing on launch, due to some currently unknown problem with GoogleSignIn


ZachOrr:
Re-opened due to https://github.com/the-blue-alliance/the-blue-alliance-ios/issues/664 - there appears to be some problem with either the latest version of GoogleSignIn, or our integration. Reverting the upgrade should give us some signal about what's wrong.

PR #755, opened 2020-01-16
Remove React Native

ZachOrr:
Let's see how much faster CI is...


ZachOrr:
Roughly 8 mins - pretty good

PR #757, opened 2020-01-16
New Event Stats views

ZachOrr:
✨ New ✨  event stats

Simulator Screen Shot - iPhone 11 Pro - 2020-01-15 at 19 54 15


the-blue-alliance-ios/View Elements/Events/Stats/EventStatsConfigurator2016.swift
```diff
@@ -0,0 +1,63 @@
+import Foundation
+import UIKit
+
+class EventStatsConfigurator2016: EventStatsConfigurator {
+
+    static func configureDataSource(_ snapshot: inout NSDiffableDataSourceSnapshot<String, InsightRow>, _ qual: [String : Any]?, _ playoff: [String : Any]?) {
+        // Match Stats
+        var matchStats: [InsightRow] = []
+
+        matchStats.append(InsightRow(title: "High Score", qual: highScoreString(qual, "high_score"), playoff: highScoreString(playoff, "high_score")))
```
ZachOrr: Let's see if we can DRY these... part of the concern is that we mismatch a key and show the wrong info

ZachOrr:
We can probably do better on DRYing up these (like just have them supply a title, key, and configuration option and do the whole table view data source thing automatically) but this has to ship so it's good enough for now

Issue #590, opened 2019-05-21
Refresh/Loading Icon sticks around after refresh/load


cyocom:
## Expected Behavior

Happens on several screens, expected is refresh/loading icon is shown until content is refreshed/loaded

## Current Behavior

Instead of the above, after the content is (re)loaded the icon sticks around.

## Screenshots

https://streamable.com/sr036

## Steps to Reproduce

1. Open app
2. Go to event with matches played
3. Go to stats tab
4. If issue doesn't happen immediately, pull down to refresh
5. Open matches tab, repeat step 4
6. Open match, repeat step 4
7. Switch to match breakdown, repeat step 4

## Your Environment

* Device: iPhone 7
* iOS version: 12.2
* TBA for iOS version: 1.1.3b1

## Context

Confusing as to why icon is still there, other than that no UX issue

cyocom:
Another reproduction case:

1. Close App
2. Go to a completed competition week
2. Go to an unloaded event
3. Go to teams
5. Select an unloaded team
6. Observe if refresh icon disappears on Summary page
7. Switch to Matches tab and observe if refresh icon disappears on matches page

https://streamable.com/5vdsj
https://streamable.com/o6pga
https://streamable.com/ideca

PR #812, opened 2020-01-28
Complete Search refresh if the user backgrounds the app

### LINKED ISSUE #811: Update search index as a background task
We should setup updating the search index as a background task that can persist across launches and update periodically in the background. Right now if a user launches TBA for iOS and we begin updating the search index, quitting the app will cause the search index to be partially complete.

ZachOrr:
Fixes #811


ZachOrr:
This should also fix some refresh weirdness in the `TeamsContianerViewController` with locking our UI thread when refreshing teams and some weirdness around updating the refresh indicator (probably related to #590)

PR #840, opened 2020-02-16
Add 2020 match breakdown

dracco1993:
## Description
Add the 2020 match breakdown

## Motivation and Context
Uh, it's kinda needed?  😅 

## How Has This Been Tested?
Only locally, but using real data from Week 0 events.

## Screenshots
Simulator Screen Shot - iPhone 11 - 2020-02-16 at 16 22 52

### DIFF
--- a/the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
+++ b/the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
@@ -125,7 +125,7 @@ struct MatchBreakdownConfigurator2020: MatchBreakdownConfigurator {
             let imageView = UIImageView(image: $0)
             imageView.autoMatch(.width, to: .height, of: imageView)
             imageView.preferredSymbolConfiguration = UIImage.SymbolConfiguration(font: UIFont.preferredFont(forTextStyle: .subheadline).bold())
-            imageView.tintColor = UIColor.black
+            imageView.tintColor = UIColor.label
             return imageView
         }, redCells).flatMap { (imgV: UIImageView, v: Int) -> [AnyHashable?] in [imgV, String(v) ] }
         
@@ -133,7 +133,7 @@ struct MatchBreakdownConfigurator2020: MatchBreakdownConfigurator {
             let imageView = UIImageView(image: $0)
             imageView.autoMatch(.width, to: .height, of: imageView)
             imageView.preferredSymbolConfiguration = UIImage.SymbolConfiguration(font: UIFont.preferredFont(forTextStyle: .subheadline).bold())
-            imageView.tintColor = UIColor.black
+            imageView.tintColor = UIColor.label
             return imageView
         }, blueCells).flatMap { (imgV: UIImageView, v: Int) -> [AnyHashable?] in [imgV, String(v) ] }
         
--- a/the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
+++ b/the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
@@ -36,14 +36,14 @@ struct MatchBreakdownConfigurator2020: MatchBreakdownConfigurator {
             rows.append(endgameRow(i: i, red: red, blue: blue))
         }
         rows.append(shieldSwitchLevelRow(title: "Shield Generator Switch Level", red: red, blue: blue))
-        rows.append(row(title: "Endgame Points", key: "", red: red, blue: blue, type: .subtotal))
-        rows.append(row(title: "Total Teleop", key: "", red: red, blue: blue, type: .total))
+        rows.append(row(title: "Endgame Points", key: "endgamePoints", red: red, blue: blue, type: .subtotal))
+        rows.append(row(title: "Total Teleop", key: "teleopPoints", red: red, blue: blue, type: .total))
         
         rows.append(row(title: "Stage Activations", key: "", red: red, blue: blue))
         rows.append(row(title: "Shield Generator Operational", key: "", red: red, blue: blue))
         
         // Match totals
-        rows.append(row(title: "Fouls / Tech Fouls", key: "foulPoints", formatString: "+%@", red: red, blue: blue))
+        rows.append(foulRow(title: "Fouls / Tech Fouls", red: red, blue: blue))
         rows.append(row(title: "Adjustments", key: "adjustPoints", red: red, blue: blue))
         rows.append(row(title: "Total Score", key: "totalPoints", red: red, blue: blue, type: .total))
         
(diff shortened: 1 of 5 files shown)


the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,258 @@
+import Foundation
+import UIKit
+
+private class BreakdownStyle2020 {
+    public static let bottomImage = UIImage(systemName: "rectangle")
+    public static let outerImage = UIImage(systemName: "hexagon")
+    public static let innerImage = UIImage(systemName: "circle")
+    public static let checkImage = UIImage(systemName: "checkmark")
```
ZachOrr: We shouldn't need to include either `checkImage` or `xImage` since we can pull them both from the shared `BreakdownStyle`

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,258 @@
...
+    public static let xImage = UIImage(systemName: "xmark")
+}
+
+struct MatchBreakdownConfigurator2020: MatchBreakdownConfigurator {
+
+    static func configureDataSource(_ snapshot: inout NSDiffableDataSourceSnapshot<String?, BreakdownRow>, _ breakdown: [String: Any]?, _ red: [String: Any]?, _ blue: [String: Any]?) {
+        
+        guard let red = red, let blue = blue else {
```
ZachOrr: nit: Everything here takes an optional, we shouldn't *need* to unwrap like this.

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,258 @@
...
+        // Add the point totals for the init line
+        guard let initLinePoints = values(key: "autoInitLinePoints", red: red, blue: blue) else {
+            return nil
+        }
+        
+        let (redLinePoints, blueLinePoints) = initLinePoints
+        redElements.append("(+\(redLinePoints ?? 0))")
+        blueElements.append("(+\(blueLinePoints ?? 0))")
+
+        return BreakdownRow(title: "Initiation Line exited", red: redElements, blue: blueElements)
```
ZachOrr: We call this `Initiation Line Extended` on web

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,258 @@
...
+                return "2"
+            } else if stage[2] == 1 {
+                return "1"
+            }
+            return ""
+        }
+        return BreakdownRow(title: title, red: [elements.first], blue: [elements.last])
+    }
+
+    private static func makeImageView(image: UIImage?, contentMode: UIView.ContentMode = UIView.ContentMode.scaleToFill) -> UIImageView {
```
ZachOrr: I love this. We should consider pulling this in to a place where it's shared. Maybe make it take a variable tint color too. We have something similar for labels in `BreakdownStyle` (Don't worry about refactoring old UIImageView creation - I can make an issue to do that)

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,258 @@
...
+
+            guard let redCellValue = rv as? Int, let blueCellValue = bv as? Int else {
+                return nil
+            }
+            redCells.append(redCellValue)
+            blueCells.append(blueCellValue)
+        }
+        
+        let redValues = zip((images).map {
+            return makeImageView(image: $0)
```
ZachOrr: We actually need to set the `contentMode` on this one to either be aspect fit or aspect fill. These are getting stretched. It's SUPER subtle - but they are. In the screenshots below, the first one is the existing image (default mode is scale to fill), and the second one is with aspect fit.

Screen Shot 2020-02-16 at 7 36 17 PM
Screen Shot 2020-02-16 at 7 36 24 PM

ZachOrr: To see if need aspect fill, test on an iPad Sim and see if things look okay if we have all 0s. If it looks weird or the images are WAY too big because they're being too greedy with their frame, we might need aspect fit or center or something.

ZachOrr: Actually here we go

## Aspect Fill
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-02-16 at 19 42 32
Simulator Screen Shot - iPhone 11 Pro - 2020-02-16 at 19 43 31

## Aspect Fit
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-02-16 at 19 46 15
Simulator Screen Shot - iPhone 11 Pro - 2020-02-16 at 19 44 35

Notice the images look pretty large even in tight situations using aspect fill, but the left-to-right spacing is a little weird. I think it's worth though to make these images a little bit more visible while maintaining the proper aspect ratio for the shapes.

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
dracco1993: The web was wrong.  😁 

the-blue-alliance/the-blue-alliance#2711

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
dracco1993: Thanks!  In 7433525, I moved it to `BreakdownStyle.imageView`, with an added optional tint, which defaults to `UIColor.label`, for easier use,

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,244 @@
...
+            let (rv, bv) = cellValues
+
+            guard let redCellValue = rv as? Int, let blueCellValue = bv as? Int else {
+                return nil
+            }
+            redCells.append(redCellValue)
+            blueCells.append(blueCellValue)
+        }
+        
+        let mode = UIView.ContentMode.scaleAspectFit;
```
ZachOrr: nit: we shouldn't need a `;` here

the-blue-alliance-ios/View Elements/Match/Breakdown/MatchBreakdownConfigurator2020.swift
```diff
@@ -0,0 +1,244 @@
...
+            return result
+        }
+        return BreakdownRow(title: title, red: elements.first ?? [], blue: elements.last ?? [])
+    }
+    
+    private static func stageActivationRow(title: String, red: [String: Any]?, blue: [String: Any]?) -> BreakdownRow? {
+        var redActivation: [Int] = [];
+        var blueActivation: [Int] = [];
+
+        for i in [3,2,1] {
```
ZachOrr: nit: Can we space these out? `[3, 2, 1]`

PR #867, opened 2020-03-05
[WIP] Only save Avatar when fetching Team Media for initial Team Info

ZachOrr:
Note to Zach: This might actually be a different problem with how we're cleaning up the media relationship. Doing a fetch in both Info and Media for team media shouldn't be a problem - one should clean up the other, and Media should fetch the missing images. The doubles might just be a problem with not maintaining the relationship properly on inserts somewhere.

Currently, going from Team Info -> Team Media will show doubles of Team Media

Simulator Screen Shot - iPhone 11 Pro - 2020-03-05 at 02 52 21

## To Do
- [ ] Clear existing Avatar when switching between years
- [ ] Fix `team.avatar` not returning a `TeamMedia` after reload/update
- [ ] Confirm this problem is reproducible when I'm less sleepy

### DIFF
--- a/Frameworks/TBAData/Sources/Team/Team.swift
+++ b/Frameworks/TBAData/Sources/Team/Team.swift
@@ -482,19 +482,42 @@ extension Team: Managed {
             $0.predicate = TeamMedia.teamYearPrediate(teamKey: key, year: year)
         }
 
-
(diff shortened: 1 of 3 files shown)


ZachOrr:
I can't seem to replicate this so I'm going to close this.

PR #870, opened 2020-03-07
Periodically refresh myTBA auth token

ZachOrr:
A pretty bad fix for the issue where our myTBA auth token is expiring after an hour and we're never refreshing it. This should periodically refresh our myTBA token every minute to ensure that users only go roughly ~60s with an expired myTBA token.

Absolutely sucks - it'll get a better fix eventually.


ZachOrr:
Kidding - we're gonna do this right

PR #899, opened 2020-10-10
Fix updating Team@Event view Event Info/Status sections crashing when Event Info section does not exist yet

### LINKED ISSUE #892: TBA 3.0.0 Beta Crashes at Teams Event View
**Describe the bug**
TBA App crashes when looking at events from the team view.

**To Reproduce**
1. Go to Teams
2. Click on any team
3. Go to events
4. Click on any event listed
5. Crash Occurs

**Expected behavior**
Should lead you to the event page of the event

**Screenshots**
Video: [Link](https://drive.google.com/file/d/17KbVmJeQhA7T4OD_WQ21FReUgbg5pN0t/view?usp=sharing)

**Your Environment (please complete the following information):**
 - Device: iPhone 11
 - OS: iOS 14.0.1
 - TBA for iOS Version: 3.0.0

ZachOrr:
Another very cool undocumented API change in iOS 14 - the [`NSDiffableDataSourceSnapshot.itemIdentifiersInSection` method](https://developer.apple.com/documentation/uikit/nsdiffabledatasourcesnapshot/3375775-itemidentifiers) no longer returns an empty array if the section doesn't exist in the snapshot, it throws an error. This PR adds a check in the Team@Event view before attempting to re-use/update the existing Event Info section Status row.

Fixes #892

### DIFF
--- a/the-blue-alliance-ios/View Controllers/Teams/Team/Team@Event/TeamSummaryViewController.swift
+++ b/the-blue-alliance-ios/View Controllers/Teams/Team/Team@Event/TeamSummaryViewController.swift
@@ -212,14 +212,15 @@ class TeamSummaryViewController: TBATableViewController {
             return nil
         }()
 
-        let eventInfoItems = snapshot.itemIdentifiers(inSection: .eventInfo)
+        let hasEventInfoSection = snapshot.indexOfSection(.eventInfo) != nil
+        let eventInfoItems = hasEventInfoSection ? snapshot.itemIdentifiers(inSection: .eventInfo) : []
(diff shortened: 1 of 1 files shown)


ZachOrr:
Note to Zach: I believe everywhere else we use this API we're doing it safely where we know the section exists before accessing objects for this section.

The [`NSDiffableDataSourceSnapshot.insertItem`](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/622ee66c3ebd33eba3f41dada6a15861b054275c/the-blue-alliance-ios/Extensions/NSDiffableDataSourceSnapshot%2BTBA.swift#L36) method is the only questionable method that will fail if we don't call `insertSection` beforehand. Luckily this is only used this exact bit of code we're modifying, and we do indeed call `insertSection` beforehand.

PR #880, opened 2020-05-23
Support Sign In with Apple

ZachOrr:
## Description
Change the myTBA Sign In view to support both Google Sign-In and Sign In with Apple

## Motivation and Context
We're required to support Sign In with Apple since we support some social logins in the app.

## How Has This Been Tested?
Manually tested

## Screenshots
Simulator Screen Shot - iPhone 11 Pro - 2020-05-23 at 12 19 17

PR #963, opened 2024-03-18
Added Match Breakdown for the 2024 game, Crescendo

dbravo1990:
## Description

Added Match Breakdown for the 2024 game, Crescendo
Based layout on the site version of the Match Breakdown 
Based functionality on previous year's standards

## Motivation and Context

There is no breakdown for 2024 yet, so I wanted to get that going before the season is over.
Will probably work on 2023 next since it's also missing :) 

## How Has This Been Tested?

Since this was only a change on the match breakdown page, I tested this against the full site on many different matches to make sure the information was correct.
I also added a couple more rows of data that aren't on the site (and let @dracco1993 know about these) that provide more data to uses about the extra RPs:
- How many notes were scored out of the total needed for a Melody RP
- How many Stage points an alliances scored for the match

## Screenshots
image

### DIFF
--- a/the-blue-alliance-ios.xcodeproj/project.pbxproj
+++ b/the-blue-alliance-ios.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 54;
+	objectVersion = 52;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


ZachOrr:
This is awesome! Thank you so much! Let me give this a look soon.

As a heads up - we're blocked right now on shipping updates to the App Store due to Apple's guidelines around needing to support Sign In with Apple. We've done the prerequisite work on the server to support this, but haven't gotten the sign in integrated into iOS yet. I'll have to take another stab at this, since https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/880 seems out of date 😅

dbravo1990:
Awesome thanks!

I would offer to provide help but that looks way over my head.

ZachOrr:
Dropped a new commit here to back out a few unrelated changes. Tests aren't passing on main right now so - going to manually merge.

Thanks for this!

Issue #948, opened 2023-03-19
Fix search UI when building against iOS 16


ZachOrr:
The text is the wrong color and there's the gap at the top that needs to be fixed.

![Simulator Screen Shot - iPhone 14 Pro - 2023-03-11 at 09 46 04](https://user-images.githubusercontent.com/516458/226147442-c8e049a8-28da-4d92-bb05-d3b4002b120c.png)

PR #987, opened 2026-04-17
Fix search UI styling and restore team avatar loading

ZachOrr:
## Summary

Four user-reported UI regressions, all small and isolated:

- **Search field legibility in light mode** — background was `darkModePrimaryBlue` (#303F9F), nearly identical to the `primaryBlue` nav bar, making the magnifying glass + placeholder disappear. New `searchFieldBackgroundColor` gives the field real contrast in both appearances, and the placeholder is now an `attributedPlaceholder` at 70% white so it stays readable.
- **Search results background mismatch** — on first focus (before typing), the empty `SearchViewController` table showed `systemGroupedBackground` (white) behind the blue search header. Results VC now paints its root view with `navigationBarTintColor`; the table keeps the grouped background so populated rows look normal.
- **Event search results had no year** — `/search_index` only returns `{key, name}`, so we parse the 4-char year prefix (same approach as the PWA). Cells now show the year on the trailing side and `[shortCode]` as a secondary line, making same-named events across seasons distinguishable. Sort order (year desc, name asc) was already correct.
- **Team avatars stopped loading** — regression from the Phase 3b Core Data → TBAAPI migration (`a3cbcf67`), which hardcoded `avatar: nil` in the header. `TeamViewController` now fetches `teamMediaByYear`, picks the `.avatar` media, and base64-decodes `details.additionalProperties["base64Image"]` into a `UIImage`. Avatar refreshes on year change; stale image is cleared first so we don't flash the previous year's avatar.

## Check out locally

```sh
git fetch origin claude/magical-nash-b3d54e
git worktree add .claude/worktrees/search-fix origin/claude/magical-nash-b3d54e
cd .claude/worktrees/search-fix
ln -s ../../../../the-blue-alliance-ios/Secrets.plist the-blue-alliance-ios/Secrets.plist
open "the-blue-alliance-ios.xcodeproj"
```

Build & run in Xcode (⌘R). When done:

```sh
cd -
git worktree remove .claude/worktrees/search-fix
```

## Test plan

- [ ] iPhone simulator, Light mode: open the Events tab — search field is readable, placeholder visible.
- [ ] iPhone simulator, Dark mode: same.
- [ ] Tap search bar with empty input — the area behind the results table matches the search header (no white band).
- [ ] Search "Kett" — each event row shows year on the right and `[miket]`-style short code under the name; results ordered newest year first.
- [ ] Open Teams → 2337 — avatar renders in the header; switching year updates the avatar instead of sticking on the old one.
- [ ] \`xcodebuild -scheme "The Blue Alliance" build\` passes (verified).


ZachOrr:
@copilot resolve the merge conflicts in this pull request

ZachOrr:
So the fix for the text color looks great - no complaints there.

Simulator Screenshot - iPhone 17 Pro - 2026-04-16 at 23 10 54

The scope segmented control is still not showing up on the initial display, so that still needs to be fixed.

Simulator Screenshot - iPhone 17 Pro - 2026-04-16 at 23 10 58

The Event search is far too cluttered. For now let's remove the event code from the subtitle text, and can we move the year into the prefix of the event name? That feels a lot cleaner to me.

Simulator Screenshot - iPhone 17 Pro - 2026-04-16 at 23 11 54

Finally, avatars do load in now, but the load in is super abrupt and jarring. Can we use some kind of animation to animate that avatar in?

https://github.com/user-attachments/assets/736bbefa-7ed7-4210-b5c7-1f23973c7c1b

The other thing that really needs fixing here is that when we initially push because we're passing a team key (instead of a Team object - even if we have a Simple Team) the header (and other information) is blank - even though we already have some information about the team. My guess is that we should update the Event and Team view controllers here to have 3 init's 

1) No previous data - pass in a team key or event key
2) Partial data like a name, number, key, etc. from Search. That gives us enough to populate some of the text fields while we're doing the initial load
3) We have a Event or Team object from the previous view and we're likely able to use that, as the model while we reload the underlying model from the API.

I think that will make these views and transitions a lot jarring and jumpy.

ZachOrr:
Also linking - this should fix #948

the-blue-alliance-ios/Protocols/SearchContainer.swift
```diff
@@ -48,13 +53,19 @@ protocol SearchContainerDelegate {
...
     }
 
-    func teamSelected(teamKey: String) {
-        let teamViewController = TeamViewController(teamKey: teamKey, dependencies: dependencies)
+    func teamSelected(teamKey: String, nickname: String?) {
+        let teamViewController = TeamViewController(teamKey: teamKey, nickname: nickname, dependencies: dependencies)
+        navigationController?.pushViewController(teamViewController, animated: true)
+    }
+
+    // Default for TeamsListViewControllerDelegate, which has the full team.
```
ZachOrr: nit: let's remove this comment - we don't really need it.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
```diff
@@ -22,6 +22,7 @@ class EventViewController: MyTBAContainerViewController, EventStatusSubscribable
 
     // MARK: - Init
 
+    // Key-only: nothing known about the event yet.
```
ZachOrr: nit: Too many comments, we don't need this.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
```diff
@@ -30,11 +31,13 @@ class EventViewController: MyTBAContainerViewController, EventStatusSubscribable
         rankingsViewController = EventRankingsViewController(eventKey: eventKey, dependencies: dependencies)
         matchesViewController = MatchesViewController(eventKey: eventKey, dependencies: dependencies)
 
+        // Raw key ("2026miket") is ugly; "2026" prefix is the best guess until the event loads.
```
ZachOrr: Don't revert my code here - use the `eventKey`.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
```diff
@@ -43,10 +46,38 @@ class EventViewController: MyTBAContainerViewController, EventStatusSubscribable
         matchesViewController.delegate = self
     }
 
+    // Partial data (e.g. key + name from search).
```
ZachOrr: nit: Too many comments here.

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
```diff
@@ -53,6 +57,22 @@ class EventInfoViewController: TBATableViewController, Refreshable, Stateful {
...
+            updateEventInfo()
+        }
+    }
+
+    func applyPartial(name: String?) {
+        partialName = name
+        if isViewLoaded, event == nil {
+            updateEventInfo()
+        }
+    }
```
ZachOrr: I really hate exposing this as external API. Part of the idea of the child view controllers is that we init them and they manage as much of their state as possible. The parent shouldn't really be managing their state like this, if we can help it. Can we attempt to refactor this?

the-blue-alliance-ios/ViewControllers/Teams/Team/TeamViewController.swift
ZachOrr, line 1: Way too many added comments in this file. Let's trim it back a notch.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr, line 1: Way too many added comments here.

the-blue-alliance-ios/ViewControllers/Teams/Team/TeamViewController.swift
ZachOrr, line 1: Right - more API I really dont' like here. I really don't like this `pendingInitialTeam`. Again - let's see what we can shift to the `EventInfoViewController`.

ZachOrr (commented):
Right - I'm attaching two videos here to outline some feedback we need to account for.

For the teams - we must've broken something with the constraints, because the year button is expanding to be way too big. I still want to have this work with dynamic text to allow users to adjust the user interface scale up, but we don't need this massively wide button. Let's make sure it has some sensible width (making sure we use modern iOS APIs for sizing). It's especially bad for teams who do not have avatars.

Image

Image

One other thing to note is that as we're pushing if the avatar is already loaded, the avatar shows up outside the bounds of the presenting view. This makes the push look really wonky.

Image

The second issue is when we're pushing to an Event and we need to load data, we basically just have a view with a single row. We don't even fill in some of the event details down below. This makes the view look really goofy - especially because it has no proper title when we push it. Let's see what we can do to load in some default UI while we're still loading the event.

https://github.com/user-attachments/assets/e18fc5b7-f6e0-43dc-b5af-d0b153f780a8

Overall this is shaping up really great! The other big thing we need to handle is the API for the Event Info VC.

the-blue-alliance-ios/Protocols/SearchContainer.swift
ZachOrr: Done in c62456fc.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr: Done in c62456fc — comment removed.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr: Already reverted in e8f73cb3 — `navigationTitle: eventKey` is back.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr: Done in c62456fc — comment removed.

the-blue-alliance-ios/ViewControllers/Teams/Team/TeamViewController.swift
ZachOrr: Trimmed in c62456fc.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr: Trimmed in c62456fc.

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
ZachOrr: Refactored in c62456fc — `EventInfoViewController.init` now takes optional `event:` and `partialName:`, and renders the detail/link sections even before the event loads so the view has structure on push. The external `apply(event:)`/`applyPartial(name:)` API is gone.

the-blue-alliance-ios/ViewControllers/Teams/Team/TeamViewController.swift
ZachOrr: Removed `pendingInitialTeam` in c62456fc — `TeamInfoViewController.init` now takes optional `team:` and seeds its own state. Parent no longer needs to hold pending data.

ZachOrr:
Addressed the review-body items in c62456fc:

- **Year button width:** content insets no longer scale with the button's frame height — fixed to 12pt horizontal. Capsule corner radius still follows height so dynamic text keeps working at larger accessibility sizes without blowing the button's width out.
- **Avatar outside bounds on push:** `clipsToBounds = true` on `TeamHeaderView`, and dropped the `UIView.animate` wrapper around the `isHidden` toggle — the image still crossfades, but the stack-view slot snaps instead of animating its collapse/expand during a push transition.
- **Empty event view on push:** `EventInfoViewController.updateEventInfo()` now always renders the detail and link sections. Alliances/Insights/Awards and Twitter/YouTube/Chief Delphi show up immediately (Twitter/YouTube/CD only need the `eventKey`, and the tap handlers for the detail rows already guarded on `event` being loaded). District Points and Website stay event-dependent.

Plus the line-level threads: all resolved — extra comments trimmed, `pendingInitialTeam` gone, and both `EventInfoViewController` / `TeamInfoViewController` now take their initial data via `init` rather than external `apply()` methods.

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
```diff
@@ -43,8 +42,10 @@ class EventInfoViewController: TBATableViewController, Refreshable, Stateful {
 
     // MARK: - Init
 
-    init(eventKey: String, dependencies: Dependencies) {
+    init(eventKey: String, event: Event? = nil, partialName: String? = nil, dependencies: Dependencies) {
```
ZachOrr: I thought we decided `partialName` wasn't the name we wanted to use here. This is returning `Event.name` from the API - can we maybe call this `name eventName: String?` (or something?) Make sure whatever we go with makes sense from a Swift API naming perspective.

ZachOrr: I also really don't like that this is a single init method. One thing I like doing with init methods is setting up multiple init methods that take some subset of data that call into some private init that takes all the data. From an API contract standpoint, it makes it where it's clear you're init-ing some view controller with some subset of functionality. I think we ought to split all the new inits that take variable subsets of information. I imagine it's okay to keep `eventKey` and `name: String? = nil` together - that's roughly the same contract of "I don't have a full Event object yet"

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
```diff
@@ -63,6 +64,8 @@ class EventInfoViewController: TBATableViewController, Refreshable, Stateful {
 
         tableView.dataSource = dataSource
         setupDataSource()
+
+        updateEventInfo()
```
ZachOrr, line 76: Do we need to explicitly call `updateEventInfo` here? I thought we had some pattern that would automatically call this. Maybe I'm wrong though - please confirm. It's been a while since I've worked on this code.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr, line 1: Yeah this is a great example of the `init` situation I was talking about in https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/987/changes#r3097837211

the-blue-alliance-ios/ViewControllers/Teams/Team/TeamInfoViewController.swift
ZachOrr, line 1: Same thing for the `init`s here - let's design them slightly better.

ZachOrr, line 1: I see this `TeamInfoViewController` has a lot of the same code I didn't like in the previous iteration of the `EventInfoViewController`. Let's refactor this one accordingly as well. Feel free to do a pass anywhere we have these external `apply` methods and get them updated more like what we did with the `EventInfoViewController`. I like the state of the `EventInfoViewController`.

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
ZachOrr: Or no - because this was what was being handled externally before? If that's the case - this is fine to leave in here! This is where I would expect it to go.

ZachOrr (commented):
Our padding changes didn't exactly work - I'm still seeing the button width jump weirdly. I also think the year button might still be too wide. We shouldn't need that much padding on the sides. I would check the iOS HIG about buttons and touch targets and go on that.

https://github.com/user-attachments/assets/5d2464ac-4cfc-469f-b794-b6f805a674ff

The code is really starting to shape up here!

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
ZachOrr: Renamed to `eventName` in bbc20b4a — matches `/search_index`'s `Event.name` field, which reads better than "partial".

the-blue-alliance-ios/ViewControllers/Events/Event/EventInfoViewController.swift
ZachOrr: Confirmed — `refresh()` only runs after the async fetch lands, so the `viewDidLoad` call is how we get the initial snapshot (seeded event / name / default rows) on first display. Leaving it in.

the-blue-alliance-ios/ViewControllers/Events/Event/EventViewController.swift
ZachOrr: Done in bbc20b4a — collapsed to two public conveniences: `init(eventKey:name:)` and `init(event:)`, both delegating to a private designated init.

the-blue-alliance-ios/ViewControllers/Teams/Team/TeamInfoViewController.swift
ZachOrr: Done in bbc20b4a — `init(teamKey:)` and `init(team:)` public conveniences + a private designated init.

ZachOrr: Done in bbc20b4a — `apply(team:)` is gone. `refresh()` now updates `self.team` + snapshot directly, and `TeamViewController.loadTeamData` no longer reaches into the child to seed state.

ZachOrr:
Took another pass on the year button in bbc20b4a:

- Migrated to `UIButton.Configuration.plain()` with `cornerStyle: .capsule` and `buttonSize: .small`. Background and highlight state now go through `config.background.backgroundColor` + `configurationUpdateHandler`, so the capsule shape and the padding come from UIKit's modern sizing rather than a manual `layer.cornerRadius` in layoutSubviews.
- Dropped the min-width constraint that was fighting intrinsic content size, and the manual leading/trailing content insets — `.small` + `.capsule` gives iOS's stock capsule sizing which is what I think you were pointing at with the HIG note.
- Switched the title font to `monospacedDigitSystemFont` — "----" and "2025" rendered at noticeably different widths before, which is probably what was reading as the button "jumping" when the year loaded in. Monospaced digits means the placeholder and the real year land at the same width, so there's no visible width change when state transitions.
- Touch target stays HIG-compliant (44pt min) via a `point(inside:)` override that expands the hit area without making the visible button bigger.

ZachOrr:
This one is maybe 90% of the way there but - gonna merge this just to get something shipped here.

PR #1011, opened 2026-04-18
fix Optional(value) showing in Event Insights page

owennkuo:
## Description
- Added conditional type casts to the highScoreRow and bonusStat functions
- Started work on the 2026 Match Breakdown (not all values showing for the moment)

## Motivation and Context
- There was a bug in the Event Insights page causing values to show Optional() around them

## How Has This Been Tested?
Tested in simulator: iOS 26.3 on iPhone 17 Pro
Tested years for event insights: 2022, 2016, 2026

## Screenshots
Before:
Screenshot 2026-04-18 at 2 18 43 PM
After:
Screenshot 2026-04-18 at 2 16 57 PM


the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift
ZachOrr, line 1: I think we should still fallback to `--` if our cast fails here just to be safe. Also is `Int` always right here? I worry sometimes we'd support floats and really we should just be trying to cast these into some `String(init:)`

ZachOrr (commented):
Good catch here! This was probably a regression with the migration from the old TBAKit -> the new API models.

Idea here makes sense to me! Feel free to re-request review once you get this buttoned up. Or - if you wanna split out the `Optional` fix from the 2026 insights - I'm good with either!

owennkuo:
Sounds good. I'll be splitting off the 2026 breakdown into a new PR.

Trying to cast the Int values into strings with 
```return "\(bonusData.safeItem(at: 0) as? String ?? "--") / \(bonusData.safeItem(at: 1) as? String ?? "--") = \(quotient)"``` 
results in "--" for all of them. I think it's fine to keep them cast as Int since match scores and opportunities should always be Int. If you'd like I can add a fallback cast as float.

the-blue-alliance-ios.xcodeproj/project.pbxproj
ZachOrr, line 1: Let's make sure we don't check in these changes - seems like something happened here. I might've messed up the package references but - let's not let that bleed in here, since we also dropped our screenshot target name.

ZachOrr (approved):
Thanks for this!

PR #1044, opened 2026-04-21
MyTBAKit for Swift 6: convert MyTBA to an actor

ZachOrr:
## Summary

Draft — first pass at getting MyTBAKit onto Swift 6 with strict concurrency. Putting this up so we can iterate together.

- **Package bumps:** `TBAUtils`, `TBATestingMocks`, `MyTBAKit` → `swift-tools-version: 6.2`, platforms `.iOS(.v26) / .macOS(.v26)` (matching TBAAPI).
- **`MyTBA` is now an actor.** Immutable-after-init state (`uuid`, `deviceName`, providers, `urlSession`) becomes `nonisolated let`. `isAuthenticated` and `fcmToken` stay `nonisolated` computed properties that read through the provider — the providers (Firebase `Auth` / `Messaging`) are the source of truth, so actor-isolating reads would add an `await` hop for zero correctness gain. The ~11 UI guard call sites (`if myTBA.isAuthenticated`) stay synchronous.
- **Auth-state surface modernized.** The old `Provider<MyTBAAuthenticationObservable>` + two-method observer protocol is replaced with a replay-on-subscribe `AsyncStream<Bool>`. Terminated continuations are pruned inline in `notifyAuthStateChanged` via `YieldResult.terminated` — no `@Sendable` closures anywhere. `Packages/TBAUtils/Sources/TBAUtils/Provider.swift` has no remaining callers and is deleted.
- **Sendable cascade** through everything crossing the actor boundary: `FCMTokenProvider`, `IDTokenProvider`, `MyTBAURLSession`, `MyTBAProtocol`, `MyTBAResponse`, `MyTBABaseResponse`, `MyTBAModel`, `MyTBAFavorite`, `MyTBASubscription`, `MyTBAPreferencesMessageResponse`, `MyTBAModelType`, `NotificationType`, `MyTBAPreferences`. All structurally safe — no `@unchecked` needed in the core package.
- **App-side conformances:** `FirebaseIDTokenProvider` is structurally Sendable (no stored state). `extension Messaging: @retroactive @unchecked Sendable {}` is the one honest `@unchecked` at the Firebase SDK boundary. The Firebase auth state listener Task-wraps `notifyAuthStateChanged` since it's now actor-isolated.
- **Observers rewritten:** `PushService`, `MyTBAViewController`, `MyTBAContainerViewController` self-subscribe via `Task { @MainActor [weak self] in let stream = await myTBA.authStateChanges(); for await ... }`, cancelled in `deinit`. AppDelegate no longer manually registers PushService as an observer.
- **`MockMyTBA` → composition.** Actors can't be subclassed, so the mock wraps a real `MyTBA` actor + exposes the test seams (`session`, `idTokenProvider`). `MyTBATestCase` keeps a `myTBA` convenience alias so most test bodies are untouched.
- **`@unchecked Sendable`** only lands in three honest places: the Firebase `Messaging` retroactive conformance, `MockURLSession`, and the mock provider/wrapper classes in the test target. Mutable-by-design test infrastructure.
- **Four new stream tests** (`emitsInitialValue`, `emitsOnChange`, `noEmitOnUnchanged`, `cleansUpOnCancel`) replacing the old Provider-based ones. All `@MainActor` so `XCTestExpectation` captures stay in-isolation.

## Test plan

- [x] `swift test --package-path Packages/MyTBAKit` — all 17 tests pass.
- [x] `xcodebuild -project the-blue-alliance-ios.xcodeproj -scheme "The Blue Alliance" -destination "generic/platform=iOS Simulator" build` — BUILD SUCCEEDED, zero concurrency warnings in the MyTBA flow.
- [x] `scripts/swift-format-lint.sh` clean.
- [ ] Manual in simulator: sign in → push registers, favorite star appears, myTBA tab refreshes. Sign out → star hides, myTBA collapses to sign-in.
- [ ] Retention check: open/close the preferences sheet ~10x; nothing leaks.

Draft — iterating together.


ZachOrr:
Gonna close for now this is getting stale.

PR #1047, opened 2026-04-21
Further CI speedups: fix cache key, wholemodule, skip indexing

ZachOrr:
**Stacked on #1045** — base this against main for now; diff will include #1045's commits until that PR lands. Once #1045 merges, this rebases cleanly to just the new commits below.

## Summary

Four targeted tweaks on top of the #1046 baseline, based on what CI run 24704426446 showed (cache restore/save keys mismatched, DerivedData cache unused, ~1,800 compile lines, 482 s in `run_tests`):

- **Cache key**: drop `**/Package.resolved` from `hashFiles`. It's gitignored, so it didn't exist at restore time but did at save time — every save got a unique key, every restore fell back to the `macOS-spm-` prefix and grabbed a stale cache.
- **Cache path**: narrow to `~/Library/Caches/org.swift.swiftpm` (shared SPM clone cache). DerivedData caching was effectively a no-op — `git checkout` resets source mtimes so Xcode rebuilt everything anyway. This keeps the useful bit (no re-clone / re-fetch of SPM deps) and drops the noise.
- **xcargs `SWIFT_COMPILATION_MODE=wholemodule`**: scoped to the CI test lane via xcargs only; local Xcode keeps incremental. CI always does clean compiles, where wholemodule is typically 20-40% faster.
- **xcargs `COMPILER_INDEX_STORE_ENABLE=NO`**: the index store powers Jump to Definition / Find Usages / Quick Help in Xcode. CI never opens Xcode, so it's pure waste.

## Test plan

- [x] Local `bundle exec fastlane test` passes with the new xcargs.
- [ ] CI `Test` job time drops vs the last consolidate-packages run (~510 s baseline).
- [ ] Warm-cache second run: `Cache SwiftPM clones` step reports a hit; `run_tests` wall-clock drops further.


ZachOrr:
Reopening against the stacked branch in my fork instead of upstream/main.

Issue #951, opened 2023-03-19
Add 2023 Event Insights


ZachOrr:
https://github.com/the-blue-alliance/the-blue-alliance-ios/wiki/Event-Insights

Example from the web https://www.thebluealliance.com/event/2023midet#event-insights

https://github.com/the-blue-alliance/the-blue-alliance/blob/acbb3f4c643e57696d3ea3c2116347d45a5ad886/src/backend/web/templates/event_partials/event_insights_2023.html

![Screenshot 2023-03-18 at 9 03 08 PM](https://user-images.githubusercontent.com/516458/226148197-092eadac-04b6-40fa-926e-270c55ae5c36.png)

Issue #841, opened 2020-02-17
Remove RP row from Match Breakdown during playoffs


ZachOrr:
We should remove the RP row from the Match Breakdown views for playoff matches. In some cases (2017) these rows are showing up as `null`

![Simulator Screen Shot - iPhone 11 Pro - 2020-02-16 at 19 27 10](https://user-images.githubusercontent.com/516458/74615903-9cba8a80-50f2-11ea-9e12-966c4ffebe1f.png)

ZachOrr:
Also a similar situation - we should make sure we're not showing any `(+ RP)` situations during for playoff matches 
![image](https://user-images.githubusercontent.com/516458/95208050-9595d900-07b6-11eb-80a5-42a4195be29d.png)

Issue #744, opened 2020-01-13
Surface Auto/Teleop Match Statistics in Event Stats


ZachOrr:
Right now we only show the overall numbers in the Event Stats, but web shows auto, teleop, and overall. We should look to surface all these numbers as well. We might consider a different interface between iPhone and iPad.

<img width="1199" alt="Screen Shot 2020-01-12 at 10 22 05 PM" src="https://user-images.githubusercontent.com/516458/72231537-07275a80-358a-11ea-9162-80c13c75dfe1.png">

https://www.thebluealliance.com/event/2019miket#event-insights

ZachOrr:
Thinking about this - we might be able to use the new collection view layout stuff to make a really killer UI that flows well on iPhone and iPad https://developer.apple.com/videos/play/wwdc2019/215/

PR #1051, opened 2026-04-21
Insights for years 2023 through 2025

owennkuo:
## Description
- [x] Add configurators for each year specifying rows and their API calls #951 
- [x] Update the view controller to use the configurators
- [x] Surface values which require a 4-column view 
- [x] Adjust previous years to surface auto/teleop/overall rather than just overall #744 
- [x] Hide RP rows if not applicable to the match in breakdown #841 

Bug Fixes:
- [x] 4 column cells selectable
- [x] Fix `filterEmptyInsights` to handle rows properly
- [x] 2026 insights where average win score is nil

## Motivation and Context
Adds support for the years that were previously missing event insights. 

## How Has This Been Tested?
Tested in simulator: iOS 26.3 on iPhone 17 Pro

## Screenshots
Screenshot 2026-04-21 at 6 42 00 PM
Screenshot 2026-04-21 at 6 42 37 PM
Screenshot 2026-04-21 at 6 42 51 PM
Screenshot 2026-04-21 at 6 43 09 PM


the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift
```diff
@@ -65,8 +65,8 @@ extension EventInsightsConfigurator {
         return InsightRow(
             title: title,
             value: .columns(
-                qual: bonusStat(qual, key) ?? [],
-                playoff: bonusStat(playoff, key) ?? []
+                qual: bonusStat(qual, key) ?? ["--", "--", "--"],
+                playoff: bonusStat(playoff, key) ?? ["--", "--", "--"]
```
ZachOrr: We should confirm our `filterEmptyInsights` method works properly with this. I'm not sure it will since now we're passing in placeholder strings instead of empty values.

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
```diff
@@ -0,0 +1,106 @@
...
+
+        bonusStats.append(
+            bonusRow(title: "Mobility", key: "mobility_count", qual: qual, playoff: playoff)
+        )
+        bonusStats.append(
+            bonusRow(title: "Auto Docked", key: "mobility_count", qual: qual, playoff: playoff)
+        )
+        bonusStats.append(
+            bonusRow(title: "Auto Engaged", key: "mobility_count", qual: qual, playoff: playoff)
+        )
```
ZachOrr, line 128:
```suggestion
        bonusStats.append(
            bonusRow(title: "Auto Docked", key: "auto_docked_count", qual: qual, playoff: playoff)
        )
        bonusStats.append(
            bonusRow(title: "Auto Engaged", key: "auto_engaged_count", qual: qual, playoff: playoff)
        )
```

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
```diff
@@ -0,0 +1,106 @@
...
+            bonusRow(
+                title: "Activation Bonus RP",
+                key: "activation_bonus_rp",
+                qual: qual,
+                playoff: nil
+            )
+        )
+        bonusStats.append(
+            bonusRow(
+                title: "\"Unicorn Matches\" (Win + Cargo Bonus + Hangar Bonus)",
```
ZachOrr: For 2023 this should be "Win + Sustainability + Activation"

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
```diff
@@ -0,0 +1,106 @@
...
+                qual: qual,
+                playoff: playoff
+            )
+        )
+        bonusStats.append(
+            bonusRow(
+                title: "Sustainability Bonus RP",
+                key: "sustainability_bonus_rp",
+                qual: qual,
+                playoff: nil
```
ZachOrr: Right we should pass `nil` for these RPs for playoffs... It actually looks like the conversion to the OpenAPI powered views broke this and for previous seasons we're now displaying `<null>` for Playoff RPs.

This is a slightly larger change but - could we modify some code here to actually hide these rows if we pass nil so we don't render them? We made that change in the web recently and I think it makes a lot of sense, and we should bring it to iOS too https://github.com/the-blue-alliance/the-blue-alliance/pull/9362

It would be great if we could update the historical match breakdowns as well, since it looks like they broke since the OpenAPI change and they render `<null>` now

Image

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
ZachOrr, line 1: I think we're missing quite a few fields we should look to add here - just to make sure these line up with the older breakdowns like 2022, 2020, etc. I know we surface less nowadays so these are easier to build out - but for 2023 we do have all the data, and we do report all the data for years like 2022 - so we should try to keep parity there.

ZachOrr (changes requested):
Thanks for this! Super excited to get these landed. 2026 seems like a good of time as any to ship the 2022 breakdowns finally 🙌 

I did have two larger architectural things here. Feel free to push back on the "remove the RP rows for playoffs" thing if you take a look and that looks like a whole thing. My thought there is like - we should be able to determine if we're passing an empty `playoffs` vs passing a `nil`, so we should use `nil` to be like "hide the row" - and that way we can get parity with the website

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
owennkuo: Screenshot 2026-04-21 at 7 28 44 PM
Would it be for this data? I think it's issue #744. Using the same 4 column cells as the bonus stats, I could set up a function for stats like those so we can surface those values.

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
ZachOrr: I fixed the nulls with https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/1053 - but removing the rows would still be good

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
ZachOrr: Right I think you're right here... ignore me for now - we can cross that bridge later. Sorry!

owennkuo: Screenshot 2026-04-21 at 9 03 56 PM
I think I've gotten it working now. Added a new function in the configurator for insightStat which accepts an array of keys (3 items). If a key is left blank, it's ignored and returns "--".

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
owennkuo: I can update the code that checks for empty rows to check only the values, not the title. After pulling from main, breakdowns aren't opening.
```Thread 1: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value```

Comment (owennkuo) [reply the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift]: For the event insights, I've switched to passing the playoff values in the rp rows. The API does have a value for this, and it's also just shown as 0 on the web.

File: the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
Comment (owennkuo) [reply the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift:1]: On another note, there are quite a few values in past years that are only showing overall values. I can add this as a task to the PR in the interest of surfacing the full details for some of these fields.

**Comment by owennkuo** [comment]
@ZachOrr couple things here.
With the `filterEmptyInsights`, the issue might be a little deeper since we're using `"--"` as the standard fallback for getting data. This could be solved by either:
- Using `nil` as the fallback instead and adding `"--"` only when we deliberately want to (e.g. playoff RP rows)
- Adding a handler for `"--"` in `filterEmptyInsights`

For hiding RP rows in playoffs, we have two cases here:
- Older years (like 2017) return `null` for any RP related rows in any playoff match
- Newer years (like 2022) return `0` for any RP related rows in any playoff match

I think the easiest way to implement this would be to add a boolean identifying a row as RP, then hiding those if the match code is identified as a playoff.
I'll do some experimenting with this to see how it would work.

**Comment by ZachOrr** [comment]
ack on the questions here - will reply when I get a chance!

**Comment by ZachOrr** [comment]
> I think the easiest way to implement this would be to add a boolean identifying a row as RP, then hiding those if the match code is identified as a playoff.

Agreed with this! Thanks for digging into the difference between the previous seasons and the modern seasons. I think this approach makes a lot of sense.

> Using nil as the fallback instead and adding "--" only when we deliberately want to (e.g. playoff RP rows)

I think this one makes the most sense to me? I'll have to look at some code once - but this one reads as the better approach of these two.

I can always do another code review and if you just leave some notes where you're uncertain I can always help architect. Sorry - I know the match breakdown code is a hairy nightmare 😓 I'll get it sorted one of these days.

File: the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift (hunk 1 above)
Comment (owennkuo) [reply the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift]: I've updated the handling logic - instead of using that as our fallback, we return an array of 3 empty strings. Row filtering will filter out rows with completely empty values for both qual and playoffs. In the view itself (after filtering), empty values are then assigned `"--"`

File: the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
Comment (ZachOrr) [reply the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift:1]: We don't need to blow this up! I'll trust your judgement here based on what we've done previously. I know previous years are spare - but I think I'm gonna rework the event insights at some point to be a little more iPad friendly. So getting them in the app right now is fine - we don't need to do it all

File: the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift (hunk 4 above)
Comment (ZachOrr) [reply the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift]: Did you get the fatal unwrap value here figured out? I can help if not - just lemme know

Comment (owennkuo) [reply the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift]: Yeah, after another merge from main everything works now. I've added a new parameter so the configurator can see the competition level, and hide the RP rows if it's anything other than `qm`.

**Comment by ZachOrr** [comment]
ack on this - just trying to finish up some web work and then I'm back to iOS work 👀

File: the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator.swift [hunk 5]
```diff
@@ -65,7 +66,7 @@ extension MatchBreakdownConfigurator {
         red: [String: Any]?,
         blue: [String: Any]?,
         type: BreakdownRow.BreakdownRowType = .normal,
-        offset: Int = 0
+        offset: Int = 0,
```
ZachOrr:
```suggestion
        offset: Int = 0
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator.swift
```diff
@@ -45,7 +46,7 @@ extension MatchBreakdownConfigurator {
...
```
ZachOrr:
```suggestion
        offset: Int = 0
```

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
```diff
@@ -0,0 +1,177 @@
...
+
+        bonusStats.append(
+            bonusRow(title: "Mobility", key: "mobility_count", qual: qual, playoff: playoff)
+        )
+        bonusStats.append(
+            bonusRow(title: "Auto Docked", key: "mobility_count", qual: qual, playoff: playoff)
+        )
+        bonusStats.append(
+            bonusRow(title: "Auto Engaged", key: "mobility_count", qual: qual, playoff: playoff)
+        )
```
ZachOrr:
```suggestion
```

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2026.swift
```diff
@@ -109,7 +109,7 @@ class EventInsightsConfigurator2026: EventInsightsConfigurator {
                 title: "Total Fuel Scored",
                 key: "total_fuel_scored",
                 qual: qual,
-                playoff: playoff
+                playoff: qual
```
ZachOrr:
```suggestion
                playoff: playoff
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2016.swift
```diff
@@ -123,8 +124,9 @@ struct MatchBreakdownConfigurator2016: MatchBreakdownConfigurator {
             row(title: "Total Score", key: "totalPoints", red: red, blue: blue, type: .total)
         )
         // RP
-        rows.append(row(title: "Ranking Points", key: "tba_rpEarned", red: red, blue: blue))
-
+        if compLevel == Optional("qm") {
```
ZachOrr: This is some code smell. I think I get this - but we should probably just do something like -
```
if let compLevel, compLevel == "qm" {
```
If `compLevel` is nil - I'm okay not showing a row here, I think. Or we can be safe and always show a `--`. But I really don't wanna wrap a string in an Optional just to satisfy a conditional like this.

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift
```diff
@@ -115,21 +129,72 @@ extension EventInsightsConfigurator {
...
         ]
     }
+    private static func insightStat(_ dict: [String: Any]?, _ key: [String]) -> [String]? {
+        guard let dict = dict else {
+            return nil
+        }
+        // Helper to fetch a value for a key at a given index safely
+        func value(for index: Int) -> Any? {
+            guard let k = key.safeItem(at: index) else { return nil }
+            return dict[k]
```
ZachOrr:
```suggestion
            guard !k.isEmpty else { return nil }
            return dict[k]
```

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsView.swift
```diff
@@ -133,6 +133,7 @@ class FourColumnTableViewCell: UITableViewCell, Reusable {
 
     var qualValues: [String] = [] {
         didSet {
+            qualValues = qualValues.map { "\($0.isEmpty ? "--" : $0)" }
```
ZachOrr: This won't be re-entrant but I'm always wary of this pattern of modifying a variable inside a didSet. I'm wary we actually want to modify the `qualValues` - I think we can probably just do something like -
```
didSet {
    let displayValues = qualValues.map { $0.isEmpty ? "--" : $0 }
    for (i, label) in qualsValueLabels.enumerated() {
        label.text = i < display.count ? displayValues[i] : ""
    }
}
```

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift
```diff
@@ -115,21 +129,72 @@ extension EventInsightsConfigurator {
...
+            }
+            if let number = value(for: 2) as? NSNumber {
+                return String(format: "%.2f", number.doubleValue)
+            }
+            guard let v = value(for: 2) else { return "" }
+            return String(describing: v)
+        }()
+
+        return [columnOne, columnTwo, columnThree]
+    }
```
ZachOrr, line 161: I think we can DRY this code quite a bit. Double check my code here - but I *think* this matches the same output you're doing here? The `let k` logic changed a little from the implementation above but 1) it still guards against empty strings and 2) although it doesn't return a nil anymore it returns a `""` which is the fallback value in the `let columnXXX` methods - so I think this is right.

```suggestion
    private static func insightStat(_ dict: [String: Any]?, _ key: [String]) -> [String]? {
        guard let dict = dict else { return nil }

        func format(_ index: Int) -> String {
            guard let k = key.safeItem(at: index), !k.isEmpty,
                let raw = dict[k] else { return "" }
            if let number = raw as? Double { return String(format: "%.2f", number) }
            if let number = raw as? NSNumber { return String(format: "%.2f", number.doubleValue) }
            return String(describing: raw)
        }

        return (0..<3).map(format)
    }
```

the-blue-alliance-ios/ViewControllers/Events/Event/Stats/EventInsightsViewController.swift
```diff
@@ -138,7 +141,7 @@ class EventInsightsViewController: TBATableViewController, Refreshable, Stateful
                     cell.qualValues = []
                     cell.playoffValues = []
                 }
-
+                cell.selectionStyle = .none
```
ZachOrr, line 144: Is this intended or accidental?

ZachOrr (commented):
This is looking great! I left suggestions for most of the diffs I think we should make here (trying to help you speed up landing this - I appreciate all the effort 🙏)

My big complaint is 1) we need to do something different than `Optional("qm")` for checking for qualification matches. Again - either we do like -
```
if let compLevel {
    if compLevel == "qm" {
        // Show RP row
    }
} else {
    // Not sure what this is - show RP row but "--"
}
```
Or we can just simplify it down to only handling the verified qualification match case (which is probably okay for what we're trying to do here)
```
if let compLevel, compLevel == "qm" {
    // Show RP row
}
```

and 2) the re-entrant bits of code in `qualValues` + `playoffValues` in `EventInsightsView`. Again - it's valid Swift, it won't cause an infinite loop - but I'd rather avoid that pattern if we can.

And then there's an errant `cell.selectionStyle = .none` in the `EventInsightsViewController.swift‎` which - that actually might be right, I'm not sure. Just confirm it's intended or revert 👍 

Thanks so much! I know it's been a lot of back and forth but this one is almost done

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator.swift
ZachOrr: nit: just remove the trailing comma here

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator.swift
ZachOrr: nit: just remove the trailing comma here

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator2023.swift
ZachOrr: These rows got fixed below but we did not remove the wrong rows - so this just removes the wrong rows

the-blue-alliance-ios/ViewControllers/Events/Event/Stats/EventInsightsViewController.swift
owennkuo: This is intended, adjusted it to match the other type of cell.

owennkuo:
@ZachOrr I think I've gotten it all. Everything is working as intended. Thanks for the review!

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift
```diff
@@ -115,21 +129,50 @@ extension EventInsightsConfigurator {
...
+        guard let dict = dict else { return nil }
+        // Helper to fetch a value for a key at a given index safely
+        func value(for index: Int) -> Any? {
+            guard let k = key.safeItem(at: index) else { return nil }
+            guard !k.isEmpty else { return nil }
+            return dict[k]
+        }
+
+        func format(_ index: Int) -> String {
+            guard let k = key.safeItem(at: index), k.isEmpty,
```
ZachOrr:
```suggestion
            guard let k = key.safeItem(at: index), !k.isEmpty,
```

the-blue-alliance-ios/ViewElements/Events/Stats/EventInsightsConfigurator.swift
```diff
@@ -115,21 +129,50 @@ extension EventInsightsConfigurator {
...
     }
+    private static func insightStat(_ dict: [String: Any]?, _ key: [String]) -> [String]? {
+        guard let dict = dict else { return nil }
+        // Helper to fetch a value for a key at a given index safely
+        func value(for index: Int) -> Any? {
+            guard let k = key.safeItem(at: index) else { return nil }
+            guard !k.isEmpty else { return nil }
+            return dict[k]
+        }
+
```
ZachOrr:
```suggestion
```

ZachOrr:
Thanks for this! 🙌

PR #1087, opened 2026-04-25
2025 Match Breakdown

owennkuo:
Add support for the 2025 Match Breakdown.

## Description
- [x] Add new configurator and rows for 2025
- [x] Create UI for viewing scoring locations

## How Has This Been Tested?
Tested in Simulator: iPhone 17 Pro on iOS 26.3, iPhone SE (3rd Generation) on iOS 26.3 (to test UI scaling)

## Screenshots
Screenshot 2026-04-26 at 5 07 31 PM
Screenshot 2026-04-26 at 5 07 38 PM

### Compared to web:
Screenshot 2026-04-26 at 5 12 37 PM


the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+        return BreakdownRow(title: title, red: elements.first ?? [], blue: elements.last ?? [])
+    }
+
+    // Coral Map Rendering
+    private static func coralMapRow(
+        title: String,
+        red: [String: Any]?,
+        blue: [String: Any]?,
+        level: Int
+    ) -> BreakdownRow? {
```
ZachOrr, line 338: We're always returning a `BreakdownRow` here even if we don't have data. I had to poke around to find a match to confirm this but - for some reason we didn't clean up [`2025cacg_f1m3`](https://www.thebluealliance.com/match/2025cacg_f1m3) and you can see in the iOS app we still render the scoring locations despite not having scoring info to show. We ought to add a check in here to make sure we have a `red`/`blue` and the data necessary to populate this.

Image

And then for example - here's [`2024cacc_f1m3`](https://www.thebluealliance.com/match/2024cacc_f1m3)

Image

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+            let legAngle = angle + (segmentIndex % 2 == 0 ? CGFloat.pi / 2 : -CGFloat.pi / 2)
+            let offset: CGFloat = 15
+
+            let offsetX = triangleCenter.x + offset * cos(legAngle)
+            let offsetY = triangleCenter.y + offset * sin(legAngle)
+            let offsetCenter = CGPoint(x: offsetX, y: offsetY)
+
+            // Rotate the coral by +15/-15 degrees based on segment index
+            if segmentIndex % 2 == 0 {
+                angle = angle - 0.261799  // Rotate even segments by -15°
```
ZachOrr: `.pi / 12` will get us a less magic number here
```suggestion
                angle = angle - .pi / 12  // Rotate even segments by -15°
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+            } else {
+                angle = angle + 0.261799  // Rotate odd segments by +15°
```
ZachOrr:
```suggestion
                angle = angle + .pi / 12  // Rotate odd segments by +15°
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+
+        let vertices: [CGPoint] = generateHexVertices(
+            centerX: center.x,
+            centerY: center.y,
+            sideLength: sideLength
+        )
+        guard let segmentIndex = letterToIndex[segmentLetter] else {
+            print("Invalid segment letter. Use A through L.")
+            return
+        }
```
ZachOrr, line 443: nit: Let's remove this print - but we can leave this as a comment
```suggestion
        // Only segment letters A through L are supported
        guard let segmentIndex = letterToIndex[segmentLetter] else {
            return
        }
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+        return path
+    }
+    private static func generateHexVertices(
+        centerX: CGFloat,
+        centerY: CGFloat,
+        sideLength: CGFloat,
+    ) -> [CGPoint] {
+        var vertices: [CGPoint] = []
+        for i in 0..<6 {
+            let angle = (CGFloat.pi / 3.0 * CGFloat(i)) + 0.523599
```
ZachOrr: Another one like the above but - `.pi / 6` gets us a slightly less magic number here. And the comment helps.
```suggestion
            let angle = (CGFloat.pi / 3.0 * CGFloat(i)) + .pi / 6  // Rotate by 30°
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+                red: red,
+                blue: blue
+            )
+        )
+
+        for i in ["auto", "coral", "barge"] {
+            rows.append(
+                bonusRankingPointRow(title: i.capitalized + " Bonus", key: i, red: red, blue: blue)
+            )
+        }
```
ZachOrr: nit: `i` is usually just when we wanna use index or some int - I think `key` here might be better
```suggestion
        for key in ["auto", "coral", "barge"] {
            rows.append(
                bonusRankingPointRow(title: key.capitalized + " Bonus", key: key, red: red, blue: blue)
            )
        }
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+            shapeLayer.frame = CGRect(x: 0, y: 0, width: width, height: height)
+            shapeLayer.fillColor = UIColor.clear.cgColor
+            shapeLayer.strokeColor = UIColor.black.cgColor
+            shapeLayer.lineWidth = 10
+            shapeLayer.path = drawHexagon(width: width, height: height).cgPath
+        }
+
+        let redContainerView = UIView(frame: CGRect(x: 0, y: 0, width: width, height: height))
+        let blueContainerView = UIView(frame: CGRect(x: 0, y: 0, width: width, height: height))
+        for i in 0..<2 {
```
ZachOrr: I think we might wanna modify this loop to be slightly simpler. I think this logic tracks? We can just setup local variables for our contianer/shape/dict/alliance in the loop and then use those - as opposed to working based on indexes

```
for (containerView, shapeLayer, dict, alliance) in [
      (redContainerView, redShapeLayer, red, Alliance.red),
      (blueContainerView, blueShapeLayer, blue, Alliance.blue),
  ] { ... }
```

the-blue-alliance-ios/Assets.xcassets/coral_image.imageset/Contents.json
```diff
@@ -0,0 +1,24 @@
...
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "properties" : {
+    "template-rendering-intent" : "template"
```
ZachOrr: Let's unset this - we don't actually wanna use `template` here.

the-blue-alliance-ios/Assets.xcassets/coral_image.imageset/Contents.json
```diff
@@ -0,0 +1,24 @@
...
+    {
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "properties" : {
```
ZachOrr, line 12: Check "Preserve Vector Data" here so iOS doesn't try to down-render our SVGs to pngs. This should help keep these images looking good at all scales.

Image

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+        }
+        func getCoralinLevel(reef: [String: Bool]?) -> [String] {
+            var coral = [String]()
+            for (key, value) in reef ?? [:] {
+                if value == true {
+                    coral.append(key)
+                }
+            }
+            return coral
+        }
```
ZachOrr:
```suggestion
        if let reef = dict?[key] as? [String: Any],
            let values = reef[levels[level - 2]] as? [String: Bool]
        {
            for (nodeKey, scored) in values where scored {
                if let id = nodeKey.last {
                    switch stage {
                    case .auto:
                        setSegment(id, to: .green, in: view, alliance: alliance, stage: .auto)
                    case .teleop:
                        setSegment(id, to: .white, in: view, alliance: alliance, stage: .teleop)
                    }
                }
            }
        }
    }
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+
+    // Coral Map Rendering
+    private static func coralMapRow(
+        title: String,
+        red: [String: Any]?,
+        blue: [String: Any]?,
+        level: Int
+    ) -> BreakdownRow? {
+        let width: CGFloat = 640
+        let height: CGFloat = 640
```
ZachOrr:
```suggestion
        let width: CGFloat = 400
        let height: CGFloat = 400
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+    }
+    private static func setSegment(
+        _ segmentLetter: Character,
+        to color: UIColor,
+        in view: UIView,
+        alliance: Alliance,
+        stage: MatchStages
+    ) {
+        let width: CGFloat = 640
+        let height: CGFloat = 640
```
ZachOrr:
```suggestion
        let width: CGFloat = 400
        let height: CGFloat = 400
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+                    ? (center.x + v1.x + mid.x) / 3 : (center.x + mid.x + v2.x) / 3,
+                y: segmentIndex % 2 == 0
+                    ? (center.y + v1.y + mid.y) / 3 : (center.y + mid.y + v2.y) / 3
+            )
+
+            var angle =
+                atan2(triangleCenter.y - center.y, triangleCenter.x - center.x) + CGFloat.pi
+
+            let legAngle = angle + (segmentIndex % 2 == 0 ? CGFloat.pi / 2 : -CGFloat.pi / 2)
+            let offset: CGFloat = 15
```
ZachOrr: related to the 640 -> 400 change
```suggestion
            let offset: CGFloat = 10
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,570 @@
...
+            // Rotate the coral by +15/-15 degrees based on segment index
+            if segmentIndex % 2 == 0 {
+                angle = angle - 0.261799  // Rotate even segments by -15°
+            } else {
+                angle = angle + 0.261799  // Rotate odd segments by +15°
+            }
+            let image = UIImage(named: "coral_image") ?? UIImage(systemName: "circle.fill")
+
+            let imageView = UIImageView(image: image?.withRenderingMode(.alwaysOriginal))
+            imageView.frame = CGRect(x: 0, y: 0, width: 125, height: 125)
```
ZachOrr: related to the 640 -> 400 change
```suggestion
            imageView.frame = CGRect(x: 0, y: 0, width: 80, height: 80)
```

ZachOrr (changes requested):
This is looking great! It's awesome to get the scoring locations added to iOS - I didn't expect that would ever get ported over. Thanks for this!

---

Let's talk briefly about the image rendering here - because it doesn't actually feel like a problem having run this on a simulator. But physical devices are different than simulators, and not all physical devices are the same. I don't think I have any lower end hardware sitting on my desk right now so - let's talk through the implications and we can try to find a way to split the difference on this one.

We're doing all the `UIGraphicsImageRenderer` work on the main thread here I believe because it's all tied to the `TableViewDataSource` cell rendering, and we want to make sure this isn't choppy for anyone. The two easiest fixes I can think of here -

1) Scale back the size we're rendering this. We're rendering at 640 and using the default graphics renderer size - which uses whatever `UIScreen.main.scale` has as a scale. So 2x for some devices and 3x for newer devices like Pro models. This is REALLY BIG for the size we're scaling it down for. Some quick math - an iPad in landscape has a 1366pt width. Our cell sizes are like roughly 25% (there's 16pt of padding). So we're around 325pt for the image size. So on iPad we're rendering this image at 640pt * 2x (screen scale) = 1280pt. But the actual size of our 325pt image view at 2x requires ~650pt. So really we're rendering this like 2x larger than we ever need it.

I dropped this down to 400 - since again, my quick back of napkin math makes me think that'll be large enough that we'll never have to upscale on any existing devices. But feel free to play around with these numbers if you try them out and they look bad.

2) I THINK we should be able to add a cache for the renderer here. If you imagine we're rendering these images for 3 rows, with 2 per side - if we render the coral for one side, I think we can cache that rendered image and just return it for the other side, or subsequent rows. And again, if we're working on the main thread for all of this we shouldn't need to worry about locking the cache or anything when doing writes. This might be more trouble than it's worth but - it would keep things fast.

Give these two things a try - but also feel free to push back if you think I'm being too much of an egghead about this. I just worry when doing graphics rendering on the main thread that we're going to hang the main thread and our app will feel choppy - and that's not what I want. I've got enough ui hitches to track down already 😅  There's also probably a tool in Instruments than can help you track this down. Time profiler isn't going to be super interesting since again - we're not running on physical hardware. But I think there's some UI rendering tool that might be helpful.

owennkuo:
Sounds good - thanks! I think I've gotten all the suggestions fixed, I'll look into the rendering cache sometime soon.

owennkuo:
@ZachOrr I've set it up to cache the `CGPoint` hexagon vertices and the hexagon `UIImage`. Does this look fine? As for hardware testing, I have an iPhone 12, which is on the lower end of our iOS 26 deployment target. However, I can't deploy directly due to provisioning profiles so it would have to be in a TestFlight build.

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,628 @@
...
+        let width: CGFloat = 400
+        let height: CGFloat = 400
+        let redShapeLayer = getCachedHexagon(width: width, height: height)
+        let blueShapeLayer = getCachedHexagon(width: width, height: height)
+
+        if red?["autoReef"] == nil || red?["teleopReef"] == nil || blue?["autoReef"] == nil
+            || blue?["teleopReef"] == nil
+        {
+            return nil
+        }
```
ZachOrr: Let's do our guard earlier here so we don't do our rendering work if we don't have to show the breakdown
```suggestion
        if red?["autoReef"] == nil || red?["teleopReef"] == nil || blue?["autoReef"] == nil
            || blue?["teleopReef"] == nil
        {
            return nil
        }

        let width: CGFloat = 400
        let height: CGFloat = 400
        let redShapeLayer = getCachedHexagon(width: width, height: height)
        let blueShapeLayer = getCachedHexagon(width: width, height: height)
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,628 @@
...
+    private static func coralMapRow(
+        title: String,
+        red: [String: Any]?,
+        blue: [String: Any]?,
+        level: Int
+    ) -> BreakdownRow? {
+        let width: CGFloat = 400
+        let height: CGFloat = 400
+        let redShapeLayer = getCachedHexagon(width: width, height: height)
+        let blueShapeLayer = getCachedHexagon(width: width, height: height)
```
ZachOrr: nit: These are images now not shape layers - we should update these variable names accordingly.

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,628 @@
...
+    ) {
+        var key = ""
+        switch stage {
+        case .auto:
+            key = "autoReef"
+        case .teleop:
+            key = "teleopReef"
+        }
+        let levels = ["botRow", "midRow", "topRow"]
+        let coral = UIImage(named: "coral_image") ?? UIImage(systemName: "circle.fill")!
```
ZachOrr: This image is going to get loaded multiple times - we ought to just be able to load these on init and use them. I don't think we need a different image instance every time, right?

the-blue-alliance-ios/Assets.xcassets/coral_image.imageset/Contents.json
```diff
@@ -0,0 +1,24 @@
...
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
```
ZachOrr: In Xcode we can set this to have a single scale so we don't have the 2x + 3x. It's Scales -> Single Scale in the asset catalog. We ought to do that for this svg 

Image

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,628 @@
...
+            ]
+            rows.append(
+                row(
+                    title: "Trough (L1) Scoring Location",
+                    key: "value",
+                    red: redL1,
+                    blue: blueL1
+                )
+            )
+        }
```
ZachOrr: I think we can clean this up a little? Double check my logic here is right I think it is. This should make it where we don't need to do those queries twice. Also keeps our nil coalescing logic slightly simpler for our 0 fallback.
```suggestion
        let redAuto = nestedValue(keys: ["autoReef", "trough"], in: red) as? Int
        let redTele = nestedValue(keys: ["teleopReef", "trough"], in: red) as? Int
        let blueAuto = nestedValue(keys: ["autoReef", "trough"], in: blue) as? Int
        let blueTele = nestedValue(keys: ["teleopReef", "trough"], in: blue) as? Int
        if (redAuto ?? redTele) != nil, (blueAuto ?? blueTele) != nil {
            rows.append(
                row(
                    title: "Trough (L1) Scoring Location",
                    key: "value",
                    red: ["value": (redAuto ?? 0) + (redTele ?? 0)],
                    blue: ["value": (blueAuto ?? 0) + (blueTele ?? 0)]
                )
            )
        }
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,628 @@
...
+            let offsetCenter = CGPoint(x: offsetX, y: offsetY)
+
+            // Rotate the coral by +15/-15 degrees based on segment index
+            if segmentIndex % 2 == 0 {
+                angle = angle - .pi / 12  // Rotate even segments by -15°
+            } else {
+                angle = angle + .pi / 12  // Rotate odd segments by +15°
+            }
+
+            let imageView = UIImageView(image: coral.withRenderingMode(.alwaysOriginal))
```
ZachOrr: nit: Our `template-rendering-intent` in our asset catalog should handle this now. I don't think we actually need to set a rendering mode here.
```suggestion
            let imageView = UIImageView(image: coral)
```

the-blue-alliance-ios/ViewElements/Match/Breakdown/MatchBreakdownConfigurator2025.swift
```diff
@@ -0,0 +1,628 @@
...
+        if let cached = hexCache.object(forKey: cacheKey) {
+            return cached
+        }
+
+        let hexagonView = UIView(frame: CGRect(x: 0, y: 0, width: width, height: height))
+        let shapeLayer = CAShapeLayer()
+        shapeLayer.frame = hexagonView.bounds
+        shapeLayer.fillColor = UIColor.clear.cgColor
+        shapeLayer.strokeColor = UIColor.black.cgColor
+        shapeLayer.lineWidth = 5
```
ZachOrr, line 587: Noting I think our lines (proportionally) got thinner. Before we had `width = 640, lineWidth = 10`, and now we have `width = 400, lineWidth = 5`. `6.25` would be the right ratio to keep our lines the same. But I'll defer to you on the visual of this. If it looks fine thinner - let's keep it thinner!

ZachOrr (changes requested):
This is looking great! We're super close here. I left some suggestions for code changes - double check they all look right, but I think so. Everything should be fairly easy so we should be able to get this landed soon! Feel free to re-request review when you're ready 👍

owennkuo:
Thank you! I think I've gotten it all. I'm gonna keep the line width at 5, as to keep it somewhat similar to the web interface.
Seems like `swift-format` isn't working right now. The linter is returning okay on my end, just not in github actions.

ZachOrr:
Bleh our lint is failing here now 🙇

ZachOrr:
Thanks for this! 🙌

ZachOrr:
FYI I'm going to ship this in a beta build - but we'll likely wait to get this to the App Store until after CMP. I don't wanna ship any unnecessary changes until after CMP if we can avoid it 🤞

PR #1091, opened 2026-04-27
Use device name when registering a device with myTBA

ZachOrr:
This one is going to stay WIP for a little. We need to get approval from Apple for the `user-assigned-device-name` entitlement. This should help users identify which device is which in their account.


ZachOrr:
Denied

PR #1113, opened 2026-05-21
Show team numbers in event Awards list

ZachOrr:
## Summary
- When the Core Data layer was removed, `AwardCellViewModel.teamDisplay` collapsed to `team.displayNickname`, which silently drops the team number whenever the team has a nickname. The Awards tab was showing just `"The Cheesy Poofs"` with no team number anywhere on the row.
- Restore the team number as the primary identifier on team-only rows (number on line 1, nickname underneath when present), matching the web's number-first emphasis while keeping the two-line touch target the cell already relies on.
- Awardee rows are unchanged in structure: awardee on line 1, bare team number on line 2 (no more `"Team N"` prefix — the repeated `Team` was noisy when stacked).

## Test plan
- [ ] Open an event with mixed award types (e.g. Smoky Mountains Regional 2026) → Awards tab
- [ ] Team-only awards (Impact, Regional Winners, etc.) show the bare team number on top, nickname underneath when one is known
- [ ] Awardee awards (Woodie Flowers, Volunteer of the Year, FIRST Leadership Finalist) show the awardee on top and the bare team number underneath
- [ ] Awards with neither team nor awardee still render `--`
- [ ] Tapping a team-only row still navigates to the team detail

### DIFF
--- a/the-blue-alliance-ios/ViewElements/Awards/AwardCellViewModel.swift
+++ b/the-blue-alliance-ios/ViewElements/Awards/AwardCellViewModel.swift
@@ -17,9 +17,12 @@ struct AwardCellViewModel {
             var awardText: [String] = []
             if let teamKey = apiRecipient.teamKey, let awardee = apiRecipient.awardee {
                 awardText.append(awardee)
-                awardText.append(Self.teamDisplay(key: teamKey, teamsByKey: teamsByKey))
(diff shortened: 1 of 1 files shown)


ZachOrr:
Fixed with https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/1114

PR #1136, opened 2026-08-28
Add setting to change app icon as well as 7 different icons

RKBoss6:
Re-added settings to change the app icon and added 7 new app icons. Changing the app icons worked perfectly in sim, but didn't work in physical device until it was updated to 26.6. My guess is a springboard caching issue, and is not consistent, so testing on other physical devices is probably necessary.

Also I changed signing in order to run it on my device, which will probably need to be changed back as I didn't want to exclude line commits from the project in case it corrupted the .xcproj.

## How Has This Been Tested?
Tested in 26.4 simulator and on physical iPhone 13 pro running 26.5

## Screenshots
Screenshot 2026-08-27 at 9 33 30 PM

IMG_3552

Screenshot 2026-08-27 at 9 34 33 PM


owennkuo:
Tried these out on my local simulator - the new icons are looking great! One note, when possible, I'd advise on using SVGs as assets in the icon composer files, for things like simple gradients or the TBA logo (if you need an SVG for that it can be found [here](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/main/the-blue-alliance-ios/AppIcon.icon/Assets/tba-lamp.svg)). Additionally, if you made changes to the signing, I'd change that back so they don't make it upstream to the main repo.

Glad you were able to get icon switching working again!


simulator_screenshot_E1F320E0-651C-40DC-87C7-B56D8513035E

RKBoss6:
Thanks! I just changed the logo to use the svg file, but for the gradients and such, I had made those in a separate software that doesn't allow for svg exporting (as far as I can tell). I think its fine to leave the smaller assets as .png, as it won't ever be seen that large in release.

I also changed the signing back to default, and I think the only thing left is to test it on physical devices, as it has sometimes failed for me until I updated my phone to 26.6.

ZachOrr:
I appreciate the work here! My diff is a little large so I wanted to call out a few things just for future reference -

1. I'm using the old `IconTableViewCell` that was in our codebase. This allows for showing images of what you're swapping to before you swap - which is nice, and how the previous app icons worked.
2. Using the cells from 1 created a bit of a headache, since the Icon Composer (`.icon`) bundles don't support just showing in an image view like the old style icons. I ended up having AI write a build script to automatically generate preview images for light + dark mode for each icon that we can use in the cells.
3. There's also code to run the build script and watch for outdated icons and generate new icons - so hopefully things will be automatic moving forward. Along with a CI check to make sure we don't check in outdated previews.

The rest of the changes are mostly just removals and cleanups. My main piece of feedback is to be mindful of what you're checking in and keeping your diff tidy! It makes it much easier to review.

I've dropped the previous season icons - partially because I've always put out year-themed icons for the year of the game, partially because we need to be mindful not to use assets we don't have the copyright for.

And then - I did a small rename of your icons to give them FRC-themed names. So Party => Champs, Sunset => Offseason. Just having a little fun with the names here.

IMG_9285
IMG_9286
IMG_9287

ZachOrr:
@RKBoss6 What's the origin of the leaf assets? Need to know where they're from to know what the licensing is and if we can redistribute them.

RKBoss6:
I made them myself in GIMP, so an external attribution isn't necessary

ZachOrr:
> I made them myself in GIMP, so an external attribution isn't necessary

Excellent!

Gonna merge this - thanks again for the work here!

PR #1178, opened 2026-09-14
Load container tabs lazily and polish empty states and list filters

ZachOrr:
Five commits, each reviewable on its own:

- **Load container tabs the first time they're shown.** `ContainerViewController` built every tab's view when a screen opened and reloaded them on each segment switch. Tabs now join the hierarchy on first selection, leaving a screen only cancels refreshes that are actually running, and the title, subtitle and bar buttons render from one `updateProperties()`.
- **Share the dependency accessors across the screen bases.** The five `api`/`myTBA`/`myTBAStores`/`statusService`/`urlOpener` accessors were copied into five classes; they come from a `DependenciesProviding` extension now. Also drops the `DataController` typealias and the unused `reloadData()` hook.
- **Commit Package.resolved in the Xcode 26 format.** Same pins. Xcode 26 rewrites it as version 3 with an `originHash`, so committing it ends the local diff and settles CI's cache key.
- **Show a title and icon on empty lists, above the keyboard.** Empty states show a tray icon, a "No Data" title and the screen's message as secondary text. The view still sits behind the list (pull-to-refresh keeps working), but its content ends at the keyboard.
- **Pin the list filters under the segmented control.** The filter moves out of the table header into a blue strip the container pins above the list, through a new `containerAccessoryView`. It's a plain `UISearchBar` now, the keyboard dismisses on drag, and team lists get a "Search Teams" placeholder.


ZachOrr:
Everything in this PR shipped in #1179, which included all of these commits.

